PR #33, opened 2015-02-08
Simple Finatra-based HTTP scoring service

avibryant:
Very simple support for loading a tree into an HTTP server and scoring features that come in via GET request.


mlmanapat:
Well we should just build this out into the multithreaded evaluator with short circuiting that I've wanted!

avibryant:
@mlmanapat well, clearly, yes. Go nuts :)

PR #34, opened 2015-02-13
Make brushfire buildable in both 2.10 and 2.11

colinmarc:
r? @avibryant

### DIFF
--- a/brushfire-core/pom.xml
+++ b/brushfire-core/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.stripe</groupId>
-  <artifactId>brushfire-core</artifactId>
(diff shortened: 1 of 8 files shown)


avibryant:
So how does one do a 2.10 build with this?

colinmarc:
You have to specify `scala.version` and `scala.binary.version`. So for example:

`mvn -Dscala.version=2.10.4 -Dscala.binary.version=2.10 clean package`

colinmarc:
This builds (and deploys) `brushfire-core` and `brushfire-scalding` just fine, but not `brushfire-finatra`, because there's no finatra for 2.11: https://github.com/twitter/finatra/pull/185

avibryant:
Ok so let's have it default to 2.10 then so it all builds, then we can have a somewhat more manual process to get the 2.11 artifacts?

colinmarc:
Makes sense. The build works now.

PR #35, opened 2015-02-14
add optional groupBy args to validate and featureImportance

avibryant:
These changes recognize that often you want to output the error or featureImportance for each of a bunch of segmentations of the validation set (for example, the error per month, for time series data).


avibryant:
cc @jocelynsarah @snoble

PR #36, opened 2015-02-18
[WIP] Prune trees to minimize validation error

roban:
Very much WIP at the moment, but it compiles, runs, and seems to to do something reasonable.

I added the digits dataset to have another example, and created `example/iris-prune` and `example/digits-prune` scripts (to run after `example/iris` or `example/digits`).

Advice welcome, @avibryant.


brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -1,6 +1,7 @@
 package com.stripe.brushfire
 
 import com.twitter.algebird._
+import scala.collection.mutable
```
avibryant: without having actually seen what you're using this for: in general using mutable collections is a red flag for me in scala

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   * @param weightMonoid
+   * @param errorOrdering
+   * @tparam P
+   * @tparam E
+   * @return
+   */
+  def prune[P, E](validationData: Map[Int, T], start: Option[Node[K, V, T]] = None, voter: Voter[T, P], error: Error[T, P, E])(implicit weightMonoid: Monoid[T], errorOrdering: Ordering[E]): Tree[K, V, T] = {
+    val begin: Node[K, V, T] = start.getOrElse(this.root) // Use root node as default root.
+    Tree(pruneNode(validationData, begin, voter, error))
+      .growByLeafIndex { i => Nil } //this will renumber all the leaves
```
avibryant: IIRC this is the second place this pattern is used, in which case maybe we should have a `renumberLeaves` method to make it more clear?

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   * @param start
+   * @param voter
+   * @param error
+   * @param weightMonoid
+   * @param errorOrdering
+   * @tparam P
+   * @tparam E
+   * @return
+   */
+  def prune[P, E](validationData: Map[Int, T], start: Option[Node[K, V, T]] = None, voter: Voter[T, P], error: Error[T, P, E])(implicit weightMonoid: Monoid[T], errorOrdering: Ordering[E]): Tree[K, V, T] = {
```
avibryant: do you expect anyone to ever use the `start` param? Why would we not just prune from the root?

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   * @param errorOrdering
+   * @tparam P
+   * @tparam E
+   * @return
+   */
+  def pruneNode[P, E](validationData: Map[Int, T], start: Node[K, V, T], voter: Voter[T, P], error: Error[T, P, E])(implicit weightMonoid: Monoid[T], errorOrdering: Ordering[E]): Node[K, V, T] = {
+    val vData = mutable.Map() ++ validationData // Create a mutable Map to allow merging of validation data as a side effect of the pruneLevel method.
+    start match {
+      case leaf: LeafNode[K, V, T] => leaf // Bounce at the bottom and start back up the tree.
+      case split: SplitNode[K, V, T] => {
```
avibryant: this would be more idiomatic as

````scala
case SplitNode[K,V,T](children) => {
````
since you only care about the `children` member, not the split as a whole.

File: brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala [hunk 5]
```diff
@@ -2,13 +2,16 @@ package com.stripe.brushfire.scalding
 
 import com.stripe.brushfire._
 import com.twitter.scalding._
+import com.twitter.algebird.AveragedValue
 
-class IrisJob(args: Args) extends TrainerJob(args) {
+abstract class csvJob(args: Args) extends TrainerJob(args) {
```
avibryant: class names should start with capital letters

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala
```diff
@@ -2,13 +2,16 @@ package com.stripe.brushfire.scalding
...
   import JsonInjections._
+  val cols: List[String]
+  implicit val stopper = FrequencyStopper[String](10, 3)
+  val error = BrierScoreError[String, Long]
+  implicit val errorOrdering = Ordering.by[AveragedValue, Double] { av => av.value } // Allows us to compare BrierScoreErrors.
 
-  val cols = List("petal-width", "petal-length", "sepal-width", "sepal-length")
-
-  val trainingData =
+  def load_data(): TypedPipe[Instance[String, Double, Map[String, Long]]] = {
```
avibryant: camelCase for methods

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -185,6 +185,40 @@ case class Trainer[K: Ordering, V, T: Monoid](
...
+        lazy val treeMap = trees.toMap
+
+        val newEx: Execution[TypedPipe[(Int, Tree[K, V, T])]] = trainingData
+          .flatMap { instance =>
+            for { // Iterate over any trees for which this instance is a validation instance.
+              (treeIndex, tree) <- treeMap if sampler.includeInValidationSet(instance.id, instance.timestamp, treeIndex);
+              leafIndex <- tree.leafIndexFor(instance.features).toList // Find the leaf that this instance falls into.
+            } yield (treeIndex, leafIndex) -> instance.target
+          } // Associate the tree and leaf indices with the instance target.
+          .sumByKey // Combine validation data for each leaf in each tree. Q: Why is this necessary?
```
avibryant: I don't think it is necessary. I think you could do `yield treeIndex -> Map(leafIndex -> target)` above and go right to the group/sum below. I'm guessing this is cargo culted off something else which maybe has the same problem?

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -185,6 +185,40 @@ case class Trainer[K: Ordering, V, T: Monoid](
...
+          .map { case ((treeIndex, leafIndex), target) => treeIndex -> Map(leafIndex -> target) } // Transform to make treeIndex the key.
+          .group // Group by tree index.
+          .withReducers(reducers)
+          .sum // Combine data for each tree into a single map.
+          .map {
+            case (treeIndex, map) => { // Iterate over each tree. The `map` is keyed by leafIndex with values being the validation data for the leaf.
+              val newTree = treeMap(treeIndex).prune(map, voter = voter, error = error) // Run the prune method on each tree, passing the validation data.
```
avibryant: the use of keyword params here seems unidiomatic to me. (In scala, unlike in python, if you get the ordering wrong the compiler will generally tell you).

example/prettydiff
```diff
@@ -0,0 +1,3 @@
+#/bin/bash
```
avibryant: did you mean to commit this file?

example/prettyprint
```diff
@@ -0,0 +1,2 @@
+#/bin/bash
```
avibryant: ditto

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala
avibryant: why is this not just `trainingData`, since that calls this and does nothing else?

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala
```diff
@@ -17,17 +20,45 @@ class IrisJob(args: Args) extends TrainerJob(args) {
...
+          _.toString
+        }.writeExecution(TypedTsv(args("output") + "/fi"))
       }
+    trainer
+  }
+
+  val trainingData = load_data()
+
+  def trainer() = {
+    args.optional("load") match {
```
avibryant: guessing you're just using this for debugging and you'll revert it back to always doing the full sequence of things before merging? (For simplicity)

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+        val newNode = SplitNode(children = split.children.map { case (k, p, n) => (k, p, pruneNode(validationData, n, voter, error)) })
+        newNode.children.exists {
+          case (k, v, s: SplitNode[K, V, T]) => true // If we have any splits as children, we can't prune.
+          case _ => false // If all children are leaves, we can prune.
+        } match {
```
avibryant: I'd find an if statement clearer than the match on true/false here

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+          case true => newNode
+          case false => pruneLevel(newNode, newNode.children.asInstanceOf[Seq[(K, Predicate[V], LeafNode[K, V, T])]], vData, voter, error)
```
avibryant: this asInstanceOf is really ugly. I think you'd be better to do something like this:

````scala
val childLeaves = newNode.children.collect{case (k,v,s: LeafNode[K,V,T]) => (k,v,s)}
if(childLeaves.size == newNode.children.size)
  pruneLevel(newNode, childLeaves, vData, voter, error)
else
  newNode
````

File: brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala [hunk 14]
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   * @param voter
+   * @param error
+   * @param weightMonoid
+   * @param errorOrdering
+   * @tparam P
+   * @tparam E
+   * @return
+   */
+  def pruneNode[P, E](validationData: Map[Int, T], start: Node[K, V, T], voter: Voter[T, P], error: Error[T, P, E])(implicit weightMonoid: Monoid[T], errorOrdering: Ordering[E]): Node[K, V, T] = {
+    val vData = mutable.Map() ++ validationData // Create a mutable Map to allow merging of validation data as a side effect of the pruneLevel method.
```
avibryant: instead of a mutable `vData` here, could you have it return a tuple with the new node and also the new validationData? That would be more idiomatic.

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   * Also merges validation data for any combined leaves. This relies on a hack that assumes no leaves have negative
+   * indicies to start out.
+   *
+   * @param parent
+   * @param children
+   * @return
+   */
+  def pruneLevel[P, E](parent: SplitNode[K, V, T], children: Seq[(K, Predicate[V], LeafNode[K, V, T])], validationData: mutable.Map[Int, T], voter: Voter[T, P], error: Error[T, P, E])(implicit weightMonoid: Monoid[T], errorOrdering: Ordering[E]): Node[K, V, T] = {
+    // Get training and validation data and validation error for each leaf.
+    val leafData: Seq[(T, T, E)] = children.map {
```
avibryant: I wouldn't normally have type annotations on any of these vals

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   * Test conditions and optionally replace parent with a new leaf that combines children.
+   *
+   * Also merges validation data for any combined leaves. This relies on a hack that assumes no leaves have negative
+   * indicies to start out.
+   *
+   * @param parent
+   * @param children
+   * @return
+   */
+  def pruneLevel[P, E](parent: SplitNode[K, V, T], children: Seq[(K, Predicate[V], LeafNode[K, V, T])], validationData: mutable.Map[Int, T], voter: Voter[T, P], error: Error[T, P, E])(implicit weightMonoid: Monoid[T], errorOrdering: Ordering[E]): Node[K, V, T] = {
```
avibryant: weightMonoid is an odd naming (maybe targetMonoid would be better?). Does that come from somewhere else?

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+    println("errors:", errors)
+    val targetSum: T = weightMonoid.sum(targets) // Combined training targets to create the training data of the potential combined node.
+    val targetPrediction: P = voter.combine(Some(targetSum))
+    val validationSum: T = weightMonoid.sum(validations) // Combined validation target for combined node.
+    println("target sum:", targetSum)
+    println("target prediction:", targetPrediction)
+    println("validation sum:", validationSum)
+    val errorOfSums: E = error.create(validationSum, targetPrediction) // Error of potential combined node.
+    val sumOfErrors: Option[E] = error.semigroup.sumOption(errors) // Sum of errors of leaves.
+    val doCombine: Boolean = sumOfErrors match {
```
avibryant: this might be better as a map followed by getOrElse

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+      case Some(soe) => errorOrdering.gt(soe, errorOfSums)
+      case None => false
+    }
+    println(s"Error of sums = ${errorOfSums} and sum of errors = ${sumOfErrors}. Should we combine? ${doCombine}")
+    println()
+    println()
+    doCombine match {
```
avibryant: if statement

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+      case false => parent
+      case true => {
+        val newIndex = -1 * children.map { case (k, p, leaf) => leaf.index }.max // Find a unique (negative) index for the new leaf.
```
avibryant: cute

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+        validationData += (newIndex -> validationSum)
+        LeafNode(newIndex, weightMonoid.sum(targets)) // Create a new leaf from the combination of the children.
```
avibryant: we already have targetSum, just use that

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +32,107 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+    val targetSum: T = weightMonoid.sum(targets) // Combined training targets to create the training data of the potential combined node.
+    val targetPrediction: P = voter.combine(Some(targetSum))
+    val validationSum: T = weightMonoid.sum(validations) // Combined validation target for combined node.
+    println("target sum:", targetSum)
+    println("target prediction:", targetPrediction)
+    println("validation sum:", validationSum)
+    val errorOfSums: E = error.create(validationSum, targetPrediction) // Error of potential combined node.
+    val sumOfErrors: Option[E] = error.semigroup.sumOption(errors) // Sum of errors of leaves.
+    val doCombine: Boolean = sumOfErrors match {
+      case Some(soe) => errorOrdering.gt(soe, errorOfSums)
```
avibryant: this should be `gte` not `gt` - if the errors are equal, we might as well combine

avibryant:
This looks correct to me, I had a bunch of style notes.

roban:
Thanks, I'll clean this up and test some more.

roban:
I'm working through the style fixes, but in the mean time I wrote some code to visualize the difference between a tree and a pruned version of the tree.

Here's an example:

image

Red nodes are the one's pruned from the tree. Green nodes are the replacement leaves.

roban:
@avibryant: thanks for all the tips. Really helpful as I get a handle on idiomatic scala. I'm going to do some testing with larger models on hadoop before I request another review.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -25,7 +25,7 @@ object Instance {
 /**
  * Produces candidate splits from the instances at a leaf node.
  * @tparam V feature values
- * @tparam T target distrubutions
+ * @tparam T target distributions
```
avibryant: thanks for catching that :)

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -31,6 +31,82 @@ case class Tree[K, V, T](root: Node[K, V, T]) {
...
+   */
+  def pruneNode[P, E](validationData: Map[Int, T], start: Node[K, V, T], voter: Voter[T, P], error: Error[T, P, E])(implicit targetMonoid: Monoid[T], errorOrdering: Ordering[E]): Node[K, V, T] = {
+    var vData = validationData
+    start match {
+      case leaf: LeafNode[K, V, T] => leaf // Bounce at the bottom and start back up the tree.
+      case SplitNode(children) => {
+        val newNode = SplitNode(children.map { case (k, p, n) => (k, p, pruneNode(vData, n, voter, error)) })
+        val childLeaves = newNode.children.collect { case (k, v, s: LeafNode[K, V, T]) => (k, v, s) }
+        if (childLeaves.size == newNode.children.size)
+          pruneLevel(newNode, childLeaves, vData, voter, error) match { case (v, n) => vData = v; n }
```
avibryant: so, if I understand this right, the vData being returned from child N is being fed into child N+1? Rather than using a local `var` for this I would prefer to see a `foldLeft` which makes that pattern more explicit.

(Sorry for all the style nits).

avibryant: wait, no. Where *does* vData get used again?

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -185,6 +185,39 @@ case class Trainer[K: Ordering, V, T: Monoid](
...
+        val newEx = trainingData
+          .flatMap { instance =>
+            for { // Iterate over any trees for which this instance is a validation instance.
+              (treeIndex, tree) <- treeMap if sampler.includeInValidationSet(instance.id, instance.timestamp, treeIndex);
+              leafIndex <- tree.leafIndexFor(instance.features).toList // Find the leaf that this instance falls into.
+            } yield treeIndex -> Map(leafIndex -> instance.target) // Yield validation instances for each tree, leaf.
+          }
+          .sumByKey // Make a single Map for each treeIndex. The Map is from leafIndex to validation target.
+          .withReducers(reducers)
+          .mapGroup {
```
avibryant: I think rather than `mapGroup` here and the `mapIter` stuff you can just `map{case (treeIndex, map) => ...}`

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -198,26 +231,28 @@ case class Trainer[K: Ordering, V, T: Monoid](
...
-                  }.values
-
-                  val featureToPermute = r.shuffle(prevInst.features).head
-                  val permuted = instance.features + featureToPermute
-
-                  val predictions = treesForInstance.flatMap { _.targetFor(permuted) }
-                  val instanceErr = error.create(instance.target, voter.combine(predictions))
-                  Some((featureToPermute._1, instanceErr))
+        val permutedFeatsPipe = trainingData
+          .map { instance: Instance[K, V, T] => r.nextDouble() -> instance } // Create a random value to sort by later.
```
avibryant: we should actually do something with a hash of the instance id

avibryant: (nice catch, though, I was lazy here)

avibryant:
This LGTM except that:
- if we're going to fix featureImportance's randomness we should do it "right" by keying off the id
- I don't understand how the merged validation data gets handled and suspect there's a bug there

avibryant:
BTW an example of what I mean by hashing properly on the id is in Samplers:
https://github.com/stripe/brushfire/blob/master/brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala#L12

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
roban: Hmm, yeah, I think you're right. So now I'm not sure why it successfully prunes multiple levels from the tree... I'll work on this.

roban: My goal was actually to pass the modified validation data back *up* the recursion stack (toward the initial call) in order to allow pruning at higher (closer to the root) levels of the tree. That kind of pruning *is* happening, but clearly this code doesn't return vData to the caller, so it can't be doing what I wanted.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
roban: `UnsortedGrouped` doesn't have a `map` method as far as I can tell, but I can do `.toTypedPipe.map`. Does that make sense?

roban:
Should I worry about the Travis error?

avibryant:
As long as it builds, no, I think I broke travis earlier (and since we don't have any tests, it doesn't seem high priority to fix :)

avibryant:
Apart from my one minor comment, LGTM

PR #37, opened 2015-03-02
Fix some scala cross-building issues

colinmarc:
Namely, brushfire-parent needs to be versioned along with everything else. I've
also added the resolve-pom-plugin, so that there aren't issues with floating
variables.

@avibryant @roban


roban:
LGTM: I successfully compiled another package against a brushfire snapshot using this patch.

PR #38, opened 2015-03-05
Initial Spark Support (Port of Scalding)

tixxit:
This isn't merge ready yet, but since I know a few other people are interested (eg @non), I figured I'd make the PR now so I can get some more eyes on it.

A few thing worth mentioning:
- Spark doesn't really have support for something like `Execution`, so we just use its strict `Unit`-happy API everywhere.
- We only write to files when explicitly asked and use `.cache()` for memoizing the trees in-between steps.
- Basically just implements `expandTimes` and `expandInMemory`, but not `prune` or `validate` (not hard, just haven't done it).

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
@@ -54,8 +54,25 @@ trait Evaluator[V, T] {
 
 /** Provides stopping conditions which guide when splits will be attempted */
 trait Stopper[T] {
+
+  /**
+   * Returns `true` if we should continue to split a leaf with the given
+   * distribution.
+   */
   def shouldSplit(target: T): Boolean
+
+  /**
+   * Returns `true` if, ideally, we should avoid splitting a leaf with the
+   * given distribution in-memory.
+   */
   def shouldSplitDistributed(target: T): Boolean
+
+  /**
+   * For the given distribution, this will return the rate at which we should
+   * sample the instances if we wish to make a split in-memory. This method
+   * should return a *safe* value, such that if we sample at the given rate,
+   * all instances should fit in memory.
+   */
   def samplingRateToSplitLocally(target: T): Double
 }
 
--- a/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
+++ b/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
@@ -1,6 +1,6 @@
-package com.stripe.brushfire.scalding
+package com.stripe.brushfire
+package scalding
 
-import com.stripe.brushfire._
 import com.twitter.scalding._
 import com.twitter.algebird._
 import com.twitter.bijection._
--- a/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
+++ b/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
@@ -32,7 +32,10 @@ case class Trainer[K: Ordering, V, T: Monoid](
 
   def execution = Execution.zip(treeExecution, unitExecution).unit
 
-  def flatMapTrees(fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[TypedPipe[(Int, Tree[K, V, T])]]) = {
+  /**
+   * Transform the decision [[Tree]]s in this [[Trainer]].
+   */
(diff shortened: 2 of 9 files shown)


tixxit:
You can run the example with something like: `SPARK_HOME=/path/to/spark ./iris-spark`

brushfire-spark/src/main/scala/com/stripe/brushfire/spark/Example.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.spark.{ SparkContext, SparkConf }
+
+class TrainCSV(
+    val argv: List[String])(
+        val cols: String*)(implicit val stopper: Stopper[Map[String, Long]] = FrequencyStopper[String](10, 3)) extends Defaults with Serializable {
+  import JsonInjections._
+
+  implicit val targetMonoid: Monoid[Map[String, Long]] = new MapMonoid[String, Long] {
+    override def sumOption(items: TraversableOnce[Map[String, Long]]): Option[Map[String, Long]] =
+      super.sumOption(items).map(m => Map.empty[String, Long] ++ m)
```
tixxit: We have to do this because Algebrid's `MutableBackedMap` isn't `Serializable`!

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -54,8 +54,25 @@ trait Evaluator[V, T] {
...
   def shouldSplit(target: T): Boolean
+
+  /**
+   * Returns `true` if, ideally, we should avoid splitting a leaf with the
+   * given distribution in-memory.
+   */
   def shouldSplitDistributed(target: T): Boolean
+
+  /**
+   * For the given distribution, this will return the rate at which we should
```
avibryant: Thanks for the comments!

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -1,6 +1,6 @@
-package com.stripe.brushfire.scalding
+package com.stripe.brushfire
+package scalding
```
avibryant: interesting, is this a standard convention? I haven't noticed it elsewhere.

brushfire-spark/src/main/scala/com/stripe/brushfire/spark/Trainer.scala
```diff
@@ -0,0 +1,226 @@
...
+    val parallelism = scala.math.min(MaxParallelism, sampler.numTrees)
+    Trainer(
+      trainingData,
+      sampler,
+      sc.parallelize(initialTrees, parallelism))
+  }
+
+  implicit class ExtraRDDOps[A](val rdd: RDD[A]) extends AnyVal {
+    def sumByKey[K, V](implicit ev: A <:< (K, V), ctK: ClassTag[K], ctV: ClassTag[V], V: Monoid[V]): RDD[(K, V)] =
+      rdd.asInstanceOf[RDD[(K, V)]].foldByKey(V.zero)(V.plus)
```
avibryant: is `foldByKey` as efficient in Spark as `reduceByKey` is? (I assume not because it seems like fold wouldn't take advantage of associative/commutative ops)

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
tixxit: It doesn't seem to be popular, but I don't think most people really think about it (especially since the package statement is usually generated). I like it because it kind of makes the relationship a bit more clear, but I don't think there is a practical difference between this vs just 1 package statement + a wildcard import.

brushfire-spark/src/main/scala/com/stripe/brushfire/spark/Trainer.scala
tixxit: Yes, it *should* be efficient. Specifically, the docs state that the "zero" may be added an arbitrary number of times to the result and cannot change it. So, it isn't `fold` in the Scala sense where the first parameter is an initial value, but rather it must be a zero in the sense of a monoid. It's kind of an unfortunate name though, since it implies inefficiency :\

brushfire-spark/src/main/scala/com/stripe/brushfire/spark/Example.scala
tixxit: Soon we can get rid of this: https://github.com/twitter/algebird/pull/424

brushfire-spark/src/main/scala/com/stripe/brushfire/spark/Trainer.scala
tixxit: I also just realized that the zero doesn't even really matter here, since the keys are sourced from `rdd`, so at least on `V` must exist anyways. I'll switch the `Monoid` to `Semigroup` and use `reduceByKey`.

PR #44, opened 2015-03-14
Use SBT instead of Maven for builds.

tixxit:
This switches the build to use SBT instead of Maven. This also fixes the current problems with the build (shade doesn't run). It may require tweaks to get publishing working (/cc @colinmarc ?). Let me know how you all feel about this/SBT.


tixxit:
Sweet - Travis completed! Just need some tests :)

README.md
```diff
@@ -44,12 +44,18 @@ cat iris.output/step_03
...
+libraryDependencies += "com.stripe" %% "brushfire" % "0.4.0"
+```
+
+To use brushfire as a jar in your own Maven project, add the following to your POM file:
 
 ```
 <dependency>
   <groupId>com.stripe</groupId>
-  <artifactId>brushfire</artifactId>
+  <artifactId>brushfire_${scala.binary.version}</artifactId>
```
colinmarc: The readme should probably just have `_2.11`, I think that's clearer. But we haven't published a scala-specific one anywhere.

colinmarc:
I had a strong negative reaction to the title of this PR - I'm not the biggest fan of SBT. But this does seem much cleaner. I guess my position would be that we should avoid the proliferation of build tools, and be consistent internally. Maven is used everywhere internally, so I understand shading/crossbuilding/publishing with it much better. 

Can you explain the issue with shading? What were we using that for?

brushfire-finatra/build.sbt
```diff
@@ -0,0 +1,20 @@
+name := "brushfire-finatra"
+
+resolvers += Resolvers.twttr
```
colinmarc: What does this line do?

tixxit: Hrmm... When I was first building, one of the deps seemed to not be on maven central, but now that I rebuild (with that line commented out) that problems seems to have disappeared. Not sure. I'll remove it now!

colinmarc: Hm, did you remove your local ivy cache before you tried to build the second time?

avibryant:
I'm in favor of this for two reasons: first, sbt is more standard/expected for scala OSS projects, so it seems like a plus to be using that here. Second, I think it's useful for us to get some experience with sbt internally in case we decide we want to switch.

colinmarc:
That makes sense to me.

tixxit:
Anyone mind if I merge?

colinmarc:
:shipit: !

PR #46, opened 2015-04-03
Add mapKeys and mapPredicates to Tree.

tixxit:
This just adds a couple of method to transform a `Tree`'s keys/values.


avibryant:
LGTM

PR #48, opened 2015-05-07
adding an in-memory driver

avibryant:
cc @tixxit 

Needs some cleanup, but provides a true local mode for brushfire (vs. relying on the local mode of scalding).


brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
```diff
@@ -0,0 +1,36 @@
+package com.stripe.brushfire.local
```
tixxit: I wonder if we should create a brushfire-examples sub-project/module too.

tixxit:
post hoc :+1:

PR #50, opened 2015-05-18
Missing features and traversal strategies

tixxit:
The goal of this PR is really to allow new ways to deal with missing features when evaluating a tree. Some changes to support this:

  * All original predicates return `true` if a feature value is missing,
  * A new predicate (`IsPresent`) was added which returns false if a feature is missing,
  * New tree traversal strategies:
    * `FirstMatch` - returns first node whose predicate is `true`,
    * `MaxWeightedMatch` - returns node whose predicate is `true` with the most weight (annotation),
    * `ProbabilisticWeightedMatch` - returns random node whose predicate is `true`, but the probability of being chosen is weighted (annotation).


brushfire-core/src/main/scala/com/stripe/brushfire/Dispatched.scala
```diff
@@ -63,17 +63,18 @@ object Dispatched {
...
         def predicates = split.predicates.map {
-          case (pred, p) => (wrapPredicate(pred)(fn), p)
+          case (pred, p) => (wrapPredicate(fn)(pred), p)
         }
       }
     }
   }
 
-  def wrapPredicate[X, A: Ordering, B, C: Ordering, D](predicate: Predicate[X])(fn: X => Dispatched[A, B, C, D]): Predicate[Dispatched[A, B, C, D]] =
+  def wrapPredicate[X, A: Ordering, B, C: Ordering, D](fn: X => Dispatched[A, B, C, D])(predicate: Predicate[X]): Predicate[Dispatched[A, B, C, D]] =
```
avibryant: I generally prefer fn arguments to go at the end (Ruby-style). I realize you're doing this to allow the currying below, but I'm not sure that's really a win. I guess we should try to come up with a lightweight style guide to address such issues.

brushfire-core/src/main/scala/com/stripe/brushfire/Predicate.scala
```diff
@@ -1,25 +1,59 @@
...
+ * A [[Predicate]] that will only return `true` if, at least, the value is
+ * defined (not missing). Normally, predicates will treat a missing value (an
+ * input of `None`) as a success and return `true`, but this is not always
+ * desired. `IsPresent` allows an additional check that *requires* the value
+ * be present to succeed.
+ *
+ * If `pred` is `None`, then this is a predicate that returns `true` iff the
+ * value is present (not missing).
+ */
+case class IsPresent[V](pred: Option[Predicate[V]]) extends Predicate[V] {
```
avibryant: I was wondering whether this is better than a simple `IsPresent` that doesn't wrap anything but that gets and-ed with another predicate ... which I guess we have to construct with de morgan's law ... which, I *think* screws us up because of the default missing feature handling? So I guess this is fine. It's certainly easier to use :)

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversalStrategy.scala
```diff
@@ -0,0 +1,149 @@
...
+import scala.math.Ordering
+import scala.util.Random
+
+/**
+ * During a tree traversal, searching for a leaf that captures a row, there
+ * are times where we have multiple successful candidate paths that can be
+ * traversed. There are many ways to choose which path to take, which are
+ * packaged up as different `TraversalStrategy`.
+ */
+sealed abstract class TraversalStrategy[K, V, T, A] {
```
avibryant: why an abstract class vs. a trait? And why seal it?

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversalStrategy.scala
```diff
@@ -0,0 +1,149 @@
...
+ * `b.annotation`, then `a` is twice as likely to be chosen (though `b` may
+ * still be chosen!).
+ *
+ * The actual path chosen depends either on `seed`, if one is provided, or the
+ * `row` itself if no `seed` is provided. So, given the same `row`, this should
+ * always result in the same leaf being found.
+ *
+ * @param seed an optional seed for the RNG used in a single traversal
+ */
+final class ProbabilisticWeightedMatch[K, V, T, A](seed: Option[Long] = None)(implicit toDouble: A => Double) extends TraversalStrategy[K, V, T, A] {
```
avibryant: I think maybe we should just require the seed. I really don't like ever relying on `hashCode` and that's too easy to do here...

avibryant:
LGTM modulu comments.

brushfire-core/src/main/scala/com/stripe/brushfire/Dispatched.scala
tixxit: I can actually just do this as an inner def instead. I usually prefer functions at the end, but saw this was really only used internally, so didn't feel to bad - but just realized it is still technically public.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversalStrategy.scala
tixxit: Oops - didn't mean to seal it!

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversalStrategy.scala
tixxit: If only we had some nice, sane `Hash` type class ;) Agreed though - I'll make the seed required. I'm a bit worried about how far to leak this in our use case, but now that I think of it, it will probably pay off in the end to expose it pretty far down.

brushfire-core/src/main/scala/com/stripe/brushfire/Predicate.scala
tixxit: Yeah - `Not` also returns `true` for missing values. I sort of saw this as the best compromise (and it is a bit easier to think about, since it's just a guard for a predicate).

PR #52, opened 2015-05-27
Add max/min target tree traversal

tixxit:
If working with something like a score, this let's us assume the worst/best case for a missing feature.


brushfire-core/src/main/scala/com/stripe/brushfire/TraversalStrategy.scala
```diff
@@ -81,6 +93,46 @@ object TraversalStrategy {
...
+   * each candidate path fully in order to find the resulting target value.
+   *
+   * Optionally, it may be possible to short-circuit some traversals if the
+   * node annotation's contain information about the minimum target in any
+   * descendant leaf nodes. This is specified by providing a `getMinTarget`
+   * function that returns a value rather than the default of `None`.
+   *
+   * @param getMinTarget function that returns the minimum target in a subtree
+   */
+  def minTargetMatch[K, V, T: Ordering, A](getMinTarget: A => Option[Min[T]] = (_: A) => None): MaxTargetMatch[K, V, T, A] =
```
avibryant: can't this just be implemented by using `maxTargetMatch` and reversing the `Ordering`?

avibryant: Er never mind. It is.

PR #53, opened 2015-05-29
Refactor tree traversals

tixxit:
This refactors tree traversals (also `TraversalStrategy` => `TreeTraversal`) to instead return a lazy Stream of matching nodes. Methods like `targetFor` on `AnnotatedTree` now require a `Semigroup[T]` to deal with multiple matching targets. The traversals can limit the number of nodes returned by calling `limitTo(n)`.

This also refactors the voters a bit (and moves them to their own file in anticipation of the training/trees split).

This would supercede #52


brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
```diff
@@ -0,0 +1,149 @@
...
+  def weightedDepthFirst[K, V, T, A: Ordering]: TreeTraversal[K, V, T, A] =
+    DepthFirstTreeTraversal(_.sortBy(_.annotation)(Ordering[A].reverse))
+
+  /**
+   * A depth-first search for matching leaves, where the candidate child leaves
+   * of a parent node are randomly shuffled, but with nodes with higher weight
+   * being given a higher probability of being ordered earlier. This is
+   * basically a mix between [[randomDepthFirst]] and [[weightedDepthFirst]].
+   */
+  def probabilisticWeightedDepthFirst[K, V, T, A <% Double](rng: Random = Random): TreeTraversal[K, V, T, A] =
```
tixxit: I kind of want to make all the `find` methods take a `Random` (with default) instead of having the traversal itself trap a random. It's a leaky abstraction, but would make things saner down the road.

tixxit:
?r @avibryant

tixxit:
Previous comment got obliterated, but I was suggesting we have `Tree#targetFor` and `TreeTraversal#find` take a `Random`, rather than have the random traversals capture the `Random` at creation. Namely, it would be a bit of a leaky abstraction, but our actual use cases wants to use a new `Random` per traversal. So the alternative means we'll be using `Random => TreeTraversal[K, V, T, A]` everywhere instead of just `TreeTraversal[K, V, T, A]` :\

tixxit:
Friendly bump =) cc @avibryant

avibryant:
@tixxit acknowledged, I'll try to look at this tomorrow.

avibryant:
LGTM in general. Re `Random`, another option to consider would be to take an `id: String` which could be used to seed the `Random` - which I think is what we actually want, and might feel slightly more general/less leaky than directly taking Random?

tixxit:
Yes - I like that more.

PR #55, opened 2015-06-17
Use Semigroup.intTimes instead of plus'ing n times.

tixxit:
Use `Semigroup.intTimes` when doing the initial sum of the `SpaceSaver`s in `SpaceSaverSplitter`.


avibryant:
LGTM.

PR #64, opened 2015-09-01
Place a Semigroup constraint on `prune` instead of all `Node` instances

NathanHowell:
This change removes the `Semigroup` constraint from the `Node` trait and places it only on methods where it is required. This will allow a greater range of annotations to be attached to leaf nodes. Users that need to prune the tree would need to transform the annotation via `mapAnnotation` to  a type with a `Semigroup` instance before pruning.

This currently slows down `TreeTraversal.weightedDepthFirst` and `TreeTraversal.probabilisticWeightedDepthFirst` because the annotation is not cached during sorting. It's not clear if these functions are normally in use. If they are I can amend the pull request to preserve sort performance.

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
@@ -5,15 +5,22 @@ import scala.util.Random
 
 import com.twitter.algebird._
 
-sealed trait Node[K, V, T, A] {
-  def annotation: A
+sealed trait Node[K, V, T, A]
+
+object Node {
+  /**
+   * Combine annotations for a tree using the [[Semigroup]] instance for [[A]]
+   */
+  def annotation[K, V, T, A: Semigroup](node: Node[K, V, T, A]): A = node match {
+    case SplitNode(children) if children.nonEmpty =>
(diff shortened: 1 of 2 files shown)


avibryant:
I defer to @tixxit on this but I'm pretty sure that caching those annotations was important (and we do indeed use them at Stripe). If you can see a way to easily preserve the sort performance that would be much preferred.

tixxit:
Yeah, caching is fairly important, since traversing a path shouldn't require traversing the whole tree to compute the annotations. But, assuming type class coherence (eg just make it clear in the docs that the annotation may be cached), I'd be OK with making the annotation an `AtomicReference`, then having the `annotation` method take a type class. Like,

```scala
  private val cachedAnnotation: AtomicReference[A] = new AtomicReference
  def annotation(implicit sg: Semigroup[A]): A = {
    var a = cachedAnnotation.get
    if (a == null) {
      cachedAnnotation.compareAndSet(null, Semigroup.sum(children.map(_._3.annotation)))
      a = cachedAnnotation.get
    }
    a
  }
```

tixxit:
Also, sorry for the delay in answering this! I think it landed mid-vacation, while I was away from my computer.

NathanHowell:
No problem, I have a few other things queued up right now but I should be back to model training in the next week or so... and should be able to work on this patch then.

PR #65, opened 2015-09-01
DispatchedSplitter should be a case class, consistent with other splitters.

NathanHowell:
This also has the benefit of inheriting from `Serialized`, which is required for Spark based training and model evaluation.


avibryant:
You should also change the creation of `DispatchedSplitter` in `Defaults.scala`, right?

NathanHowell:
I've updated it, as well as a few other formatting tweaks.

NathanHowell:
With case classes the generated `apply` method on the companion object just calls `new`, so it wasn't actually needed but is good for consistency.

avibryant:
Thanks!

PR #68, opened 2015-09-03
Introduce a Double splitter based on TDigest's cumulative density estimates

NathanHowell:
This pull request adds a splitter that relies on `TDigest` to provide target distribution estimates at fixed quantile intervals. It currently has no tests, I just wanted to get this up to get a review started.

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
@@ -26,15 +26,29 @@ private [this] case object TDigestSemigroup extends Semigroup[TDigest] {
 }
 
 object TDigestSplitter {
-  private def splitDists[L](q: Double)(target: L, digest: TDigest): (Map[L, Long], Map[L, Long]) = {
+  /**
+   * Estimate the number of items on either side of a quantile split
+   */
+  private def splitCounts(q: Double, digest: TDigest): (Long, Long) = {
     // the cumulative density reported is sometimes outside of [0,1] so we need to bound it. otherwise
     // the estimated target distribution will be far from realistic... leading the Evaluator to
     // frequently pick a suboptimal split point
-    val leftDist = (digest.cdf(q).max(0.0).min(1.0) * digest.size().toDouble).toLong
-    val rightDist = (digest.size - leftDist).max(0L)
-    val left = if (leftDist > 0L) Map(target -> leftDist) else Map.empty[L, Long]
-    val right = if (rightDist > 0L) Map(target -> rightDist) else Map.empty[L, Long]
+    val left = (digest.cdf(q).max(0.0).min(1.0) * digest.size().toDouble).toLong
+    val right = (digest.size - left).max(0L)
     (left, right)
+  }
+
+  /**
+   * Create a singleton [[scala.collection.Map]] if the value is positive, else return [[scala.collection.Map.empty]]
+   */
+  private def positiveOrEmpty[L]: (L, Long) => Map[L, Long] = {
+    case (key, value) if value > 0L => Map(key -> value)
+    case _ => Map.empty[L, Long]
+  }
+
+  private def targetDistribution[L](q: Double)(target: L, digest: TDigest): (Map[L, Long], Map[L, Long]) = {
+    val (left, right) = splitCounts(q, digest)
+    (positiveOrEmpty(target, left), positiveOrEmpty(target, right))
   }
 }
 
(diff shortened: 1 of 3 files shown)


brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    implicit val tds = TDigestSemigroup
+    val splits = for {
+      merged <- Semigroup.sumOption(stats.valuesIterator).toSeq
+      i <- 1 to k
+      q = merged.quantile(i.toDouble / k.toDouble).max(0.0).min(merged.size().toDouble)
+      (leftDist, rightDist) <- Semigroup.sumOption(stats.map(Function.tupled(splitDists(q))))
+      if leftDist.nonEmpty || rightDist.nonEmpty
+    } yield {
+      BinarySplit(LessThan(q), leftDist, Group.minus(parent, rightDist))
```
NathanHowell: I've wondered about the `Group.minus` bit here. Many of the splitters (maybe all aside from `QTree`) do it, so I'm wondering if this is used by anyone or if this should also be done in `QTreeSplitter`?

avibryant:
Hm so `S` should just be `Map[L,TDigest]` (for a `Splitter[Double, Map[L,Long]]`). 

What's going on here is that a `T` is meant to be a distribution of the label `L` (in this case, by simply counting how many instances there are of each value of `L`). And the splitter's `S` needs to be a joint distribution of `Double` and `L` - so in principle you want to be counting co-occurences of a given `Double` value and a given `L` value. But the `TDigest` is already a distribution of `Double`, so what you want is the `Map[L,TDigest]`.

What you have probably works because the inputs are `Map[L,Long]` where the `Long` is always `1L` (because for any given single instance, that's what the label "distribution" looks like). And so that just becomes your somewhat unwieldy representation of a given `L` value. But it wouldn't work if you provided training data with different weights for some instances, and the semantics are all wonky.

NathanHowell:
Took a stab at defining `type S = Map[L, TDigest]`, which also seems to work for some simple tests. Is this going in the right direction?

NathanHowell:
Fixed up the type variable names to be `L` instead of `T` and removed some other dumb stuff. Think this probably covers your feedback now.

brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  override def semigroup: Semigroup[S] = {
+    implicit val tds = TDigestSemigroup
+    implicitly[Semigroup[S]]
+  }
+
+  override def create(value: Double, target: Map[L, Long]): S = {
+    target.mapValues {
+      case count if count <= Int.MaxValue.toLong =>
```
avibryant: hm so in the unlikely event that we have a count > Int.MaxValue we just silently drop it? I feel like this might be somewhere that throwing an exception is warranted.

NathanHowell: It's a partial function that will thrown an exception if no case statements
match. I can certainly throw a more specific exception though.

brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
```diff
@@ -0,0 +1,78 @@
...
+
+    val splits = for {
+      // merge the statistics from all targets
+      merged <- Semigroup.sumOption(stats.valuesIterator).toSeq
+      i <- 1 to k
+      q = merged.quantile(i.toDouble / k.toDouble).max(0.0).min(merged.size().toDouble)
+      (leftDist, rightDist) = Monoid.sum(stats.map(Function.tupled(splitDists(q))))
+      if leftDist.nonEmpty || rightDist.nonEmpty
+    } yield {
+      BinarySplit(LessThan(q), leftDist, rightDist)
```
avibryant: You had asked earlier about the `Group.minus` thing. I think it's fine to ignore for now but let me explain: the joint distribution, `S`, is only going to include instances who have a value for the feature we're splitting on. But there may be some instances that would have made it to the node we're splitting, but *don't* have that feature. We have the distribution for the parent node, so if we want, and if our target is a `Group`, we can figure out how many instances must be missing this feature by doing the subtraction (and choose to include them on one or the other sides of the split).

However, that's effectively hard-coding the missing-feature traversal strategy during training (something we *don't* do anymore in evaluation). So I'm not even sure if it's the right thing, or if we should default to just dropping them.

The one place we *need* it is in the sparse boolean splitter, where (for efficiency) we expect a one-hot encoding where ie the feature only appears if at all if it's true and the falses are all missing.

brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
```diff
@@ -0,0 +1,78 @@
...
+
+      case (f@Some(acc), el) =>
+        acc.add(el)
+        f
+    }
+  }
+}
+
+object TDigestSplitter {
+  private def splitDists[L](q: Double)(target: L, digest: TDigest): (Map[L, Long], Map[L, Long]) = {
```
avibryant: nit, and not necessary for merge, but it seems slightly better to have this function just return `(Long,Long)` and do the conversion to `Map[L,Long]` separately (and reusably)?

NathanHowell: I'll do this, would be good for performance.

brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
NathanHowell: The other option would be to change it to be a `Map[L, Int]` instead... it's definitely more correct, but it doesn't compose with a bunch of other classes that assume `Map[L, Long]`

avibryant: Yeah, this seems right for now. Probably hints at something better we should be doing in the long run with those bunch of other classes.

brushfire-core/src/main/scala/com/stripe/brushfire/TDigest.scala
NathanHowell: Actually... is this possible without grouping by `target: T` first?

avibryant: yes, you'd call this for each target, but you don't need to pass the target in our return it, the caller can handle that. (Really not a big deal, just a slight simplification of this method).

NathanHowell: Current PR has this logic factored out to a new function

avibryant:
Sorry, missed the further commits until just now. LGTM - thanks!

PR #69, opened 2015-09-22
Local mode fixes

avibryant:
These are necessary to get decent performance out of the local mode.

- Tree's in-memory `expand` wasn't taking the sampler or tree id, so it couldn't correctly do feature sampling. This led to overly-correlated trees and poor performance.
- The local `validate` method was still generating (empty) predictions for instances that had been completely filtered out of the validation set (eg by the `OutOfTimeSampler`). This made performance look worse than it actually was.


tixxit:
:+1:

PR #70, opened 2015-09-22
Compute AUC when validating

avibryant:
And, more generally, get a list of ConfusionMatrix objects, one for each threshold, giving access to all the standard binary classification metrics.


brushfire-core/src/main/scala/com/stripe/brushfire/Errors.scala
```diff
@@ -36,6 +36,38 @@ case class BrierScoreError[L, M](implicit num: Numeric[M])
   }
 }
 
+case class ConfusionMatrix(
```
tixxit: Mind if we put this in its own file?

PR #72, opened 2015-09-25
sbt assembly improvements

NathanHowell:
This PR bumps sbt to the latest in the 0.13 branch and fixes three minor issues:

* The top-level assembly goal failed because `brushfire-core` does not pull in `MakeJar.settings` and thus fails to merge due to duplicate classes in minlog and kryo. I've disabled the assembly plugin for `brushfire-core` since it's not terribly useful.
* Avoids a sbt deprecation warning by changing `jarName` to `assemblyJarName`.
* Marks the hadoop related bits as `provided`, reducing the assembly jar size a bit.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -24,7 +24,8 @@ lazy val root = project.
   settings(unpublished: _*)
 
 lazy val brushfireCore = project.
-  in(file("brushfire-core"))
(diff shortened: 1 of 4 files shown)


avibryant:
Thanks!

PR #74, opened 2015-10-09
Annotated instances WIP

NathanHowell:
NOTE: this doesn't work right now, wanted to open this now before I actually complete the work.

I'm looking at efficient ways to calculate some statistics from Instance instances and attach them to tree leaf nodes. One idea I had was to replace the timestamp parameter (which I don't always need) with `A` and use `Semigroup.sumOption` to populate `LeafNode` instances as they're created. If the timestamp is needed it may also be an interesting thing to track with a `(Min[Long], Max[Long])` tuple (or something fancier) to see time ranges per node.

The real world problem is that I have a lot of `Instance`s and a lot of customers. An `Instance` is tied to a single customer, but a customer can have multiple `Instance`s. I want to count the number of distinct customers per split (via `Monoid[HyperLogLog]`, which is usually much different from the target distribution. It would be interesting to calculate these statistics per target as well.


tixxit:
I feel like something like `Aggregator` may be better suited here to "sum" instance annotations. Namely, the annotation on the instance (call it `L`) is not necessarily the same type that we want as the final tree node annotation (call it `A`). So, `Trainer` would also take some `Aggregator[L, _, A]`. In your case, your `L` is the customer and `A` would be an HLL.

NathanHowell:
Yeah, that would work just fine. Seems like this would also require removing the `Semigroup` constraint on `A` (related to that other open pull request). Do you think this approach is worth pursuing?

NathanHowell:
Related question: would the cached `annotation` on `SplitNode` still be necessary for your usage?

avibryant:
Ah interesting. For a while we had a `weight: W` attribute on `Instance` for nearly exactly this use case (multiple charges per merchant, interested in unique merchants per split). What we switched to was just having a more complex `T` value - like, instead of having `T` be `Map[Boolean,Long]` it could be `Map[Boolean,(Long,HLL)]`. Would that work for you? There's probably some worthwhile work to be done making this kind of thing more convenient...

Incidentally, a strategy we've played with for the problem you *may* be worrying about here (of overfitting to specific, high-volume customers), is to use the customer for the Instance `id`, which means trees are sampling whole customers instead of individual instances, which means that no customer can be in every tree, which helps with the overfitting. (This, too, might be worth making more convenient as a pattern).

NathanHowell:
It's for two different problems: one is preventing high traffic users from dominating the trees (the top 1% are being discarded entirely now), the other is for visualization and pattern/sequence mining.

Sticking a HLL in `T` would also be acceptable. I'll try this out later today.

NathanHowell:
Another related idea: splits that cover a wider time period might be preferred over a narrower period, particularly towards the root of a tree.

avibryant:
Yeah, interesting idea.

It does feel unsatisfying to combine the simple minimization of training error (or information gain or however you want to think about it) with what are basically anti-overfitting heuristics (ranging from a simple minimum instance count to this kind of much more sophisticated thing). It does seem reasonable for the framework to deal with these independently...

I wonder if the right thing here is to have a `Instance[M,Map[K,V],T]`, where: 
* we roll the idea of timestamp and id into a generic metadata type param
* `Sampler` operates on `M`
* we have some new trait which provides `M => N` and `Semigroup[N]` for aggregations of the metadata to annotate the nodes with
* we expect splits to be evaluated both on `T` and on `N`

NathanHowell:
Okay, now I'm attempting to implement your last suggestion. It's also not functional yet but it does seem a bit nicer. Does this seem like it's going in the right direction?

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
```diff
@@ -60,40 +60,44 @@ case class RFSampler(numTrees: Int, featureRate: Double, samplingRate: Double =
...
   }
 
-  def includeFeature(name: K, treeIndex: Int, leafIndex: Int) = base.includeFeature(name, treeIndex, leafIndex)
+  def includeFeature(metadata: M, name: K, treeIndex: Int, leafIndex: Int) = {
+    base.includeFeature(metadata, name, treeIndex, leafIndex)
+  }
 }
 
-case class OutOfTimeSampler[K](base: Sampler[K], threshold: Long) extends Sampler[K] {
+case class OutOfTimeSampler[-M, -K](timestamp: M => Long, base: Sampler[M, K], threshold: Long) extends Sampler[M, K] {
```
NathanHowell: I could make this implicit and add a view from `DefaultMetadata => Long`, or leave as-is.

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -26,14 +36,15 @@ object Tree {
         val edges = split.predicates.toList.map {
           case (pred, _) =>
             val newInstances = instances.filter { inst => pred(inst.features.get(splitFeature)) }
+            val annotation = Semigroup.sumOption(instances.map { inst => annotator.create(inst.metadata) })
```
NathanHowell: Is there a non-empty list somewhere in algebird with a monoid instance? .. or should I just use `.get`?

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -8,11 +8,21 @@ object Tree {
   def apply[K, V, T](node: Node[K, V, T, Unit]): Tree[K, V, T] =
     AnnotatedTree(node)
 
-  def singleton[K, V, T](t: T): Tree[K, V, T] = Tree(LeafNode(0, t, ()))
+  def singleton[K, V, T, A: Semigroup](t: T, a: A): AnnotatedTree[K, V, T, A] = {
```
NathanHowell: I'd still like to drop the semigroup requirement here

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala
```diff
@@ -32,8 +32,8 @@ abstract class CSVJob(args: Args) extends TrainerJob(args) {
     fn
   }
 
-  def train(trainingData: TypedPipe[Instance[String, Double, Map[String, Long]]]): Trainer[String, Double, Map[String, Long]] = {
-    Trainer(trainingData, KFoldSampler(4))
+  def train(trainingData: TypedPipe[Instance[DefaultMetadata, Map[String, Double], Map[String, Long]]]): Trainer[DefaultMetadata, Unit, String, Double, Map[String, Long]] = {
+    Trainer[DefaultMetadata, Unit, String, Double, Map[String, Long]](trainingData, KFoldSampler[DefaultMetadata](id = _.id, 4))
```
NathanHowell: Type signature seems to be required here, need to find a workaround

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -89,7 +90,7 @@ case class Trainer[K: Ordering, V, T: Monoid](
             case (treeIndex, map) => {
               val newTree =
                 treeMap(treeIndex)
-                  .updateByLeafIndex { index => map.get(index).map { t => LeafNode(index, t) } }
+                  .updateByLeafIndex { index => map.get(index).map { t => LeafNode(index, t, Monoid.zero[A]) } }
```
NathanHowell: Shouldn't be using `Monoid.zero[A]`

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -151,10 +152,10 @@ case class Trainer[K: Ordering, V, T: Monoid](
               val newTree =
                 treeMap(treeIndex)
                   .growByLeafIndex { index =>
-                    for (
-                      (feature, split, _) <- map.get(index).toList;
+                    for {
+                      (feature, split, _) <- map.get(index).toList
                       (predicate, target) <- split.predicates
-                    ) yield (feature, predicate, target, ())
+                    } yield (feature, predicate, target, Monoid.zero[A])
```
NathanHowell: Shouldn't be using `Monoid.zero[A]`

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -222,7 +223,7 @@ case class Trainer[K: Ordering, V, T: Monoid](
    * @tparam E
    * @return
    */
-  def featureImportance[P, E](error: Error[T, P, E])(fn: TypedPipe[(K, E)] => Execution[_])(implicit voter: Voter[T, P]) = {
+  def featureImportance[P, E](error: Error[T, P, E])(fn: TypedPipe[(K, E)] => Execution[A])(implicit voter: Voter[T, P]) = {
```
NathanHowell: `Execution[A]` fixes the typechecker but feels a bit odd. I'm guessing something else is broken.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -293,13 +294,13 @@ case class Trainer[K: Ordering, V, T: Monoid](
             .map {
               case ((treeIndex, leafIndex), instances) =>
                 val target = Monoid.sum(instances.map { _.target })
-                val leaf = LeafNode[K, V, T, Unit](0, target)
-                val expanded = Tree.expand(times, leaf, splitter, evaluator, stopper, instances)
+                val leaf = LeafNode[K, V, T, A](0, target, Monoid.zero[A])
```
NathanHowell: Shouldn't be using `Monoid.zero[A]`

avibryant:
Oh great! This definitely feels like the right direction. (Will respond to your more detailed comments later).

avibryant:
(At this point do we want to make `AnnotatedTree` be just `Tree`? I don't think we use it without an `A` anywhere anymore, right?)

NathanHowell:
`AnnotatedTree` isn't required anymore. The latest push removes this, but is still not fully working yet.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -5,21 +5,30 @@ import com.twitter.algebird._
...
  * @constructor create a new instance
- * @param id an identifier unique to this instance
- * @param timestamp the time this instance was observed
+ * @param metadata metadata about this instance
  * @param features a map of named features that make up this instance
  * @param target a distribution of predictions or labels for this instance
  */
-case class Instance[K, V, T](id: String, timestamp: Long, features: Map[K, V], target: T)
+case class Instance[M, F, T](metadata: M, features: F, target: T) {
+  def id(implicit eq: M =:= DefaultMetadata): String = {
```
avibryant: I understand adding these for source-level compatibility, but I think I'd rather just make this a breaking change (we can bump the version number appropriately) and get rid of the special casing here - or rather, push it to the places that actually need it.

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
```diff
@@ -2,18 +2,18 @@ package com.stripe.brushfire
...
-  def timesInTrainingSet(id: String, timestamp: Long, treeIndex: Int) = 1
-  def includeInValidationSet(id: String, timestamp: Long, treeIndex: Int) = true
-  def includeFeature(name: Any, treeIndex: Int, leafIndex: Int) = true
+  def timesInTrainingSet(metadata: Any, treeIndex: Int) = 1
+  def includeInValidationSet(metadata: Any, treeIndex: Int) = true
+  def includeFeature(metadata: Any, name: Any, treeIndex: Int, leafIndex: Int) = true
 }
 
-case class KFoldSampler(numTrees: Int) extends Sampler[Any] {
+case class KFoldSampler[-M](id: M => String, numTrees: Int) extends Sampler[M, Any] {
```
avibryant: I wonder if we want `id: M=>String` here or if it's better to have some kind of `M:Hashable` which builds in the murmur hashing. (Unfortunately Algebird has never really standardized a Hashable typeclass...)

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
avibryant: Or this could change to being a `MetadataFilterSampler` or something that takes a `M=>Boolean` rather than hardcoding the idea of a `Long` threshold.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
tixxit: As a side note, I've set up sbt-release and released a new version of Diorama. We plan on being much better about this in the future!

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
tixxit: Adding a return type may help suss out why the `_` isn't working.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
tixxit: Do we need the annotations during training? May be easier to fix `A` to `Unit` or something and just avoid the whole thing. If this doesn't work with what you are doing, then no worries... we'll have to figure something out.

tixxit: AH - nvm, I think you are using it...

avibryant: Yeah I don't think it's in this PR but definitely the idea is to use metadata to inform splits. (Which probably means adapting `Stopper`, I think).

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
NathanHowell: I'll yank this out.

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
NathanHowell: There is a dependency on `com.twitter:util-hashing` already, which looks like it might work.

NathanHowell: Erp, it's only used by the finatra server...

NathanHowell:
It's getting closer, still some failing tests though.

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
NathanHowell: Opinions? The current revision uses `util-hashing`... not sure if it's much better than using `M => String` since the other samplers have similar arguments for extracting timestamps and the like.

brushfire-core/src/main/scala/com/stripe/brushfire/package.scala
```diff
@@ -1,6 +1,5 @@
 package com.stripe
 
 package object brushfire {
-  type Tree[K, V, T] = AnnotatedTree[K, V, T, Unit]
```
NathanHowell: Should I just delete this file?

avibryant: :+1:

avibryant:
You asked about Hashable but I can't find that comment here anymore. Anyway, I kinda agree that `M => String` is probably just as good for now.

NathanHowell:
Alright, I've backed out the `Hashable` goo and tried to minimize the diff.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -320,18 +344,18 @@ case class Trainer[K: Ordering, V, T: Monoid](
   }
 
   /** add out of time validation */
-  def outOfTime(quantile: Double = 0.8) = {
+  def outOfTime(quantile: Double = 0.8)(implicit eq: M =:= DefaultMetadata): Trainer[M, K, V, T, A] = {
```
NathanHowell: Probably better to make this `timestamp: M => Long`

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -339,19 +363,26 @@ case class Trainer[K: Ordering, V, T: Monoid](
...
   val MaxReducers = 20
 
-  def apply[K: Ordering, V, T: Monoid](trainingData: TypedPipe[Instance[K, V, T]], sampler: Sampler[K]): Trainer[K, V, T] = {
-    val empty = 0.until(sampler.numTrees).map { treeIndex => (treeIndex, Tree.singleton[K, V, T](Monoid.zero)) }
+  def apply[M, K: Ordering, V, T: Monoid, A](
+      trainingData: TypedPipe[Instance[M, Map[K, V], T]],
+      annotator: Annotator[M, A],
+      sampler: Sampler[M, K],
+      id: M => String): Trainer[M, K, V, T, A] = {
+    implicit val annotationSemigroup = annotator.monoid
```
NathanHowell: Seems a bit arbitrary to have the `Annotator` instance carry around the `Monoid[A]` (excuse the variable name here) but `Monoid[T]` comes from a constraint

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala
```diff
@@ -32,8 +32,17 @@ abstract class CSVJob(args: Args) extends TrainerJob(args) {
...
   }
 
-  def train(trainingData: TypedPipe[Instance[String, Double, Map[String, Long]]]): Trainer[String, Double, Map[String, Long]] = {
-    Trainer(trainingData, KFoldSampler(4))
+  private def id(m: DefaultMetadata): String = {
+    m.id
+  }
+
+  def train(trainingData: TypedPipe[Instance[DefaultMetadata, Map[String, Double], Map[String, Long]]]): Trainer[DefaultMetadata, String, Double, Map[String, Long], Unit] = {
+    val annotator = defaultMetadataAnnotator[Unit]
```
NathanHowell: `Annotator` isn't an implicit so it needs to be passed explicitly... so it's not really a "default"

brushfire-finatra/src/main/scala/com/stripe/brushfire/finatra/BrushfireServer.scala
```diff
@@ -7,7 +7,7 @@ import com.twitter.finagle.http._
 import com.twitter.bijection._
 
 class BrushfireServer extends FinatraServer {
-  def score[K, V, T: Semigroup, P](root: String, trees: Iterable[Tree[K, V, T]], voter: Voter[T, P])(fn: ParamMap => Map[K, V]) = {
+  def score[K, V, T: Semigroup, P](root: String, trees: Iterable[Tree[K, V, T, Unit]], voter: Voter[T, P])(fn: ParamMap => Map[K, V]) = {
```
NathanHowell: If the annotations aren't used by the server, a `Unit` annotation will parse successfully but will use less memory than `A` (whatever it is) at runtime... question is: are annotations useful for a model server?

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -2,38 +2,334 @@ package com.stripe.brushfire
...
+        })
+
+      case LeafNode(index, target, annotation) =>
+        LeafNode(index, target, annotation)
+    }
+
+    Tree(recur(root))
+  }
+
+  private def mapLeaves[T0, A0: Semigroup](f: (T, A) => (T0, A0)): Tree[K, V, T0, A0] = {
```
NathanHowell: Might be a bit closer to a `fold` than a `map` now that `SplitNode` sums up the child annotations...

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
```diff
@@ -60,40 +60,44 @@ case class RFSampler(numTrees: Int, featureRate: Double, samplingRate: Double =
...
+    if (includeInTraining(metadata))
+      base.timesInTrainingSet(metadata, treeIndex)
     else
       0
   }
 
-  def includeInValidationSet(id: String, timestamp: Long, treeIndex: Int) = {
-    timestamp >= threshold
+  def includeInValidationSet(metadata: M, treeIndex: Int) = {
+    !includeInTraining(metadata)
```
NathanHowell: Should this have a second predicate for `includeInValidationSet`? Or is simply negating the `includeInTraining` result sufficient?

brushfire-core/src/main/scala/com/stripe/brushfire/Lazy.scala
```diff
@@ -0,0 +1,36 @@
+package com.stripe.brushfire
+
+import com.twitter.algebird.{Monoid, Semigroup}
+
+/**
+  * Wrapper class to allow expensive [[Semigroup]] operations on a [[Tree]] to be deferred until first use
+  */
+class Lazy[A](fn: () => A) extends Serializable {
```
NathanHowell: Seems like this might be better to include in algebird directly? Doesn't seem to be a good fit in here, but it does provide compatibility with the way `SplitNode` used to behave for it's `lazy val annotation: A` field.

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
```diff
@@ -135,58 +157,70 @@ object JsonInjections {
...
-          Option(n.get("leaf")) match {
-            case Some(indexNode) => fromJsonNode[Int](indexNode).flatMap { index =>
-              fromJsonNode[T](n.get("distribution")).map { target =>
-                LeafNode(index, target)
+        private def invertSplit(n: JsonNode): Try[SplitNode[K, V, T, A]] = {
+          import Monad.scalaTry // needed for Applicative.sequence
+          for {
+            annotation <- fromJsonNode[A](n.get("annotation"))
+            splitsNode <- tryChild(n, "splits")
+            splits <- Applicative.sequence {
```
NathanHowell: I haven't tested this stuff (yet), just wondering if this amount of Haskell-ish code is alright? The `children.find(_.failure)/children.map(_.get)` bit was equivalent to `Applicative.sequence/Try.recoverWith`.

NathanHowell:
Alright, done with comments for now.

NathanHowell:
Following up to @tixxit's earlier comment: using `Aggregator` may be the best route. We're half way there already and I end up needing something like `Aggregator.present` in our training pipeline.

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
avibryant: Seems pretty reasonable to me. I've never looked at algebird's Monad stuff before, thanks for giving me an excuse to :)

brushfire-core/src/main/scala/com/stripe/brushfire/Lazy.scala
avibryant: Agreed on getting this into Algebird. I'm actually kinda surprised it doesn't exist there already.

brushfire-core/src/main/scala/com/stripe/brushfire/Samplers.scala
avibryant: I can't offhand think of a use case where you would want anything other than this single predicate. Given that, though, I wonder if a better name is `MetadataHoldoutSampler` (or just `HoldoutSampler`?)

brushfire-finatra/src/main/scala/com/stripe/brushfire/finatra/BrushfireServer.scala
avibryant: We use annotations in a model server to guide traversals (eg for how you deal with missing features, either organically missing or deleted as part of some feature importance scheme). In general I'm pretty sure that people will find creative uses for them.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
avibryant: Agreed.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
avibryant: Yeah, I mean - the only reason for that is there's no other natural place for the Monoid[T] to live... I agree it's pretty arbitrary but I can live with it.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Example.scala
avibryant: Should it not go in `Defaults` then, but maybe be `DefaultMetadata.annotator` instead?

avibryant:
Hm, if I understand @tixxit's comment, though, he's implying that the final (post-present) type is what would live in the actual tree? I guess that's ok as long as you are storing those values for the interior nodes as well, but the fact that you lose the ability to sum the metadata for two subtrees to get the metadata for the parent bothers me a bit...

avibryant:
(OTOH from a serialization POV I get it: it's often going to be *much* easier to serialize the final presentation type than the intermediate, summable one. So on reflection that probably does make it worthwhile).

avibryant:
@NathanHowell what's the status of this PR - should we consider it complete at this point?

NathanHowell:
I have some fixes from Spark that need to get rolled into the Scalding
driver. Otherwise it's working.

PR #76, opened 2015-12-10
Make all splits binary

striation:
This PR simplifies brushfire's conception of split nodes (and splits) so that all decisions are binary. In practice this is almost always the case.

Some benefits of this:

1. Should drastically reduce the number of `Predicate` instances we need.
2. Avoid allocating a collection instances per split node.
3. Simplify algorithms which had to encode/decode implicit binary structure.
4. Nicer to use single concrete `Split` class.

Some drawback:

1. Different serialization format.
2. Source/binary-incompatible change.

What do you all think?

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
@@ -2,24 +2,37 @@ package com.stripe.brushfire
 
 import com.twitter.algebird._
 
-sealed trait Node[K, V, T, A] {
+import java.lang.Math.{abs, max}
+
+sealed abstract class Node[K, V, T, A] {
   def annotation: A
+
+  def renumber(nextId: Int): (Int, Node[K, V, T, A]) =
+    this match {
+      case SplitNode(p, k, lc0, rc0, a) =>
+        val (n1, lc) = lc0.renumber(nextId)
+        val (n2, rc) = rc0.renumber(n1)
+        (n2, SplitNode(p, k, lc, rc, a))
+      case LeafNode(_, target, annotation) =>
+        (nextId + 1, LeafNode(nextId, target, annotation))
+    }
 }
 
-case class SplitNode[K, V, T, A: Semigroup](children: Seq[(K, Predicate[V], Node[K, V, T, A])]) extends Node[K, V, T, A] {
-  require(children.nonEmpty)
-
-  lazy val annotation: A =
-    Semigroup.sumOption(children.map(_._3.annotation)).get
-
-  def findChildren(row: Map[K, V]): List[Node[K, V, T, A]] =
-    children.collect { case (k, p, n) if p(row.get(k)) => n }(collection.breakOut)
+case class SplitNode[K, V, T, A](predicate: Predicate[V], key: K, leftChild: Node[K, V, T, A], rightChild: Node[K, V, T, A], annotation: A) extends Node[K, V, T, A] {
+  def evaluate(row: Map[K, V]): List[Node[K, V, T, A]] =
+    predicate(row.get(key)) match {
+      case Some(true) => leftChild :: Nil
+      case Some(false) => rightChild :: Nil
+      case None => leftChild :: rightChild :: Nil
+    }
 }
 
-case class LeafNode[K, V, T, A](
-  index: Int,
-  target: T,
-  annotation: A) extends Node[K, V, T, A]
+object SplitNode {
+  def apply[K, V, T, A: Semigroup](p: Predicate[V], k: K, lc: Node[K, V, T, A], rc: Node[K, V, T, A]): SplitNode[K, V, T, A] =
+    SplitNode(p, k, lc, rc, Semigroup.plus(lc.annotation, rc.annotation))
+}
+
+case class LeafNode[K, V, T, A](index: Int, target: T, annotation: A) extends Node[K, V, T, A]
 
 object LeafNode {
   def apply[K, V, T, A: Monoid](index: Int, target: T): LeafNode[K, V, T, A] =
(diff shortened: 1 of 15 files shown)


brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
```diff
@@ -2,24 +2,37 @@ package com.stripe.brushfire
...
+      case Some(true) => leftChild :: Nil
+      case Some(false) => rightChild :: Nil
+      case None => leftChild :: rightChild :: Nil
+    }
+}
 
-  def findChildren(row: Map[K, V]): List[Node[K, V, T, A]] =
-    children.collect { case (k, p, n) if p(row.get(k)) => n }(collection.breakOut)
+object SplitNode {
+  def apply[K, V, T, A: Semigroup](p: Predicate[V], k: K, lc: Node[K, V, T, A], rc: Node[K, V, T, A]): SplitNode[K, V, T, A] =
```
tixxit: Super nit-picky, but could we put the key first?

striation: Sure, I can make that change.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -42,14 +42,23 @@ trait Splitter[V, T] {
 }
 
 /** Candidate split for a tree node */
-trait Split[V, T] {
-  def predicates: Iterable[(Predicate[V], T)]
+case class Split[V, T](predicate: Predicate[V], leftDistribution: T, rightDistribution: T) {
```
tixxit: So nice having this just be a case class.

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
```diff
@@ -124,57 +124,49 @@ object JsonInjections {
...
+            obj.put("left", toJsonNode(lc)(nodeJsonNodeInjection))
+            obj.put("right", toJsonNode(rc)(nodeJsonNodeInjection))
+            obj
         }
 
-        def tryChild(node: JsonNode, property: String) = Try {
+        def tryChild(node: JsonNode, property: String): Try[JsonNode] =
+          Try {
+            val child = node.get(property)
+            assert(child != null, property + " != null")
```
tixxit: I know this existed before, but could we make this `require` instead? It is reasonable to expect that we'll be given bad JSON, where as `assert`s are more meant for sanity checks that may possibly be elided.

striation: Oh yeah. I used `require` in the new code, I'll change this too.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
```diff
@@ -148,10 +148,10 @@ case class DepthFirstTreeTraversal[K, V, T, A](order: (Random, List[Node[K, V, T
           Stream.empty
         case (leaf @ LeafNode(_, _, _)) :: rest =>
           leaf #:: loop0(rest)
-        case (split @ SplitNode(_)) :: rest =>
-          val newStack = split.findChildren(row) match {
+        case (split @ SplitNode(_, _, _, _, _)) :: rest =>
+          val newStack = split.evaluate(row) match {
             case Nil => rest
-            case node :: Nil => node :: rest
+            //case node :: Nil => node :: rest
```
tixxit: This is here to avoid calling `order` in the common case. Since there is only 1 order for a singleton list, we know what the result would be for any `order` function (which is not allowed to change the size).  Mainly, our `order` functions are not always very, uh, speedy, so if we could avoid them with a simple `if`, that is desirable.

tixxit:
Re: serialization format - is there an easy way to include some backwards compatibility here, assuming all our splits are actually just binary splits? Otherwise this will make the migration path quite a bit harder, as we have production models using the old format.

tixxit:
@NathanHowell : Mind taking a look?

tixxit:
Overall I'm really happy with this change. Great stuff!

avibryant:
@tixxit I suspect we can pretty easily write a script (in Scala or even Ruby or something) that translates the JSON.

tixxit:
@avibryant The problem is that we need to support both formats in production for a short time. Upgrading the affected models should be relatively easy.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
striation: OK, I'll take a look! ;)

tixxit:
@NathanHowell Will the serialization format change make your life much harder? If not, then we can just support the old format from within our model server for the change over, then delete the code.

striation:
Yeah I imagined writing a script to update our models. And like @tixxit says, I think writing a custom injection internally that can parse either format (temporarily) is a way to switch over our services to the new code without breaking anything.

NathanHowell:
Is it possible to ease the use of binary splits without completely removing the ability to have n-way splits?

I've been working on implementing an MDLP splitter for TDigest (based on http://ijcai.org/Past%20Proceedings/IJCAI-93-VOL2/PDF/022.pdf) that will generate multiple splits per node. The goal would be to reduce tree depth (and training iterations, important for very large training sets) without affecting model quality. It is not universally applicable though. Orthogonal trees/forests may similarly benefit from multiway splits.

striation:
@NathanHowell First of all, sorry for opening a PR removing a feature you need.

Just out of curiosity, is there a problem with encoding a mult-way split as a series of binary splits? The memory/performance overhead of this should be pretty much the same as it was under the previous implementation which used a collection of triples. I don't think it should make the trees any bigger, as far as allocations or size in memory goes.

It would be easy to provide a nice constructor to make instantiating a "logical" multi-way split simpler. I could also imagine displaying a tree of many splits on the same feature as a multi-way split logically (sort of the dual of the previously mentioned constructor). Is there something else that would be needed?

I think I can definitely support an explicit multi-way split if needed but I want to be sure I understand the requirement first.

NathanHowell:
@striation it could certainly be a tree of splits instead of a single split, that is fine. I was just imaging that there would be implications for tree expansion to fill in targets (and possibly annotations) for each split point, even if they're not a leaf.

avibryant:
This could be a follow-up PR, but my proposal is that we get rid of `Split` and instead have `Splitter` directly produce `Node`s, which could then be a series of binary splits to simulate a multi-way. We then need a way for `Evaluator` to evaluate a `Node`; my proposal is that its interface changes to take `Iterable[T]`, which is just the `T`s living in all the leaves of a `Node`, regardless of how deep the tree might be.

Incidentally this makes it very clear that `Evaluator` should actually be called `TrainingError`, and could usefully be applied to a whole tree...

avibryant:
We have three big pull requests in flight - this one, https://github.com/stripe/brushfire/pull/74, and now https://github.com/stripe/brushfire/pull/77 . @NathanHowell @striation any thoughts on the right order to try to land them in?

Issue #51, opened 2015-05-18
separate brushfire-tree from brushfire-training


avibryant:
Some applications, like model servers, will not need the training code. It would be nice if they could depend on a smaller subset of the code. Likely, this would include:

- Tree
- Predicate
- Voter
- the serialization Injections
- Dispatched

but hopefully would not include:

- Instance
- Splitter
- Error
- Sampler

tixxit:
Something to think about while doing this, is adding a `case class Forest(trees: Map[Int, Tree[K, V, T])` in `brushfire-training` while we're at it. A lot of helper methods could be added to this to make the creation of `Trainer`s easier.

avibryant:
`Forest` is probably useful in `brushfire-trees` too, right? Then I guess we could have an implicit `ForestTrainingHelpers`...

PR #77, opened 2015-12-21
WIP: TrainingStep

avibryant:
attn @tixxit @non

The intent here is to capture the basic mechanics of various training steps - updateTargets, expand, prune, etc - in a way that can be reused in multiple execution environments (local, scalding, spark, ...).

For now, this is all added directly to and used only by the Local trainer. The near-term impact is that the local trainer will stream over its input data in the same way that the distributed trainers do, rather than requiring it to all be loaded into memory. For this PR to be complete, we should move the training steps to their own module (maybe also doing https://github.com/stripe/brushfire/issues/51), and refactor the scalding trainer to use them.

It's very possible that this is too much or too little abstraction - right now it seems a bit overfit to the needs of the specific training steps and platforms we support, and I suspect it will be brittle going forward. (In fact, featureImportance already doesn't work with this, though I can argue that we should move to a TreeTraversal-based strategy for that which would). At the same time, I think *some* approach like this will be valuable going forward, and I think it's better to start going imperfectly down this path.

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
@@ -57,7 +57,7 @@ case class Expand[K,V,T](sampler: Sampler[K], stopper: Stopper[T], splitter: Spl
     for
       ((treeIndex, tree) <- trees.toList;
       i <- 1.to(sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex)).toList;
-      leaf <- tree.leafFor(instance.features).toList if stopper.shouldSplit(leaf.target) && stopper.shouldSplitDistributed(leaf.target);
+      leaf <- tree.leafFor(instance.features).toList if stopper.shouldSplit(leaf.target);
       (feature, stats) <- features if (sampler.includeFeature(feature, treeIndex, leaf.index)))
         yield (treeIndex, leaf.index, feature) -> stats
   }
@@ -90,12 +90,35 @@ case class Expand[K,V,T](sampler: Sampler[K], stopper: Stopper[T], splitter: Spl
   }
 }
 
+
+trait OutputStep[K,V,T,O] {
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]): Seq[O]
+  def semigroup: Semigroup[O]
+}
+
+case class Validate[K,V,T:Semigroup,P,E](sampler: Sampler[K], error: Error[T, P,E], voter: Voter[T, P])
+  extends OutputStep[K,V,T,E] {
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]) = {
+    val predictions =
+      for (
+        (treeIndex, tree) <- trees
+          if sampler.includeInValidationSet(instance.id, instance.timestamp, treeIndex);
+        target <- tree.targetFor(instance.features).toList
+      ) yield target
+
+    List(error.create(instance.target, voter.combine(predictions)))
+  }
+
+  val semigroup = error.semigroup
+}
+
 case class Trainer[K: Ordering, V, T: Monoid](
     trainingData: Iterable[Instance[K, V, T]],
     sampler: Sampler[K],
     trees: List[Tree[K, V, T]]) {
 
-  def updateTrainer(step: TrainingStep[K,V,T]): Trainer[K,V,T] = {
+  def trainingStep(step: TrainingStep[K,V,T]): Trainer[K,V,T] = {
     val treeMap = trees.zipWithIndex.map{case (t,i) => i->t}.toMap
     var sums1 = Map[(Int,Int,step.K1),step.V1]()
 
@@ -111,7 +134,9 @@ case class Trainer[K: Ordering, V, T: Monoid](
 
     var sums2 = treeMap.mapValues{tree => Map[Int, step.V2]()}
     sums1.foreach{case ((treeIndex, leafIndex, k), v1) =>
-      step.lift(treeMap(treeIndex), leafIndex, k, v1).foreach{v2 =>
+      val lifted = step.lift(treeMap(treeIndex), leafIndex, k, v1)
+
+      lifted.foreach{v2 =>
         val combined = sums2(treeIndex).get(leafIndex) match {
           case Some(old) => step.semigroup2.plus(old,v2)
           case none => v2
@@ -129,38 +154,25 @@ case class Trainer[K: Ordering, V, T: Monoid](
     Trainer(trainingData, sampler, newTrees)
   }
 
-  def updateTargets =
-    updateTrainer(UpdateTargets(sampler))
-
-  def expand(implicit splitter: Splitter[V, T], evaluator: Evaluator[V, T], stopper: Stopper[T]) =
-    updateTrainer(Expand(sampler, stopper, splitter, evaluator))
-
-/*
-  def prune[P, E](error: Error[T, P, E])(implicit voter: Voter[T, P], ord: Ordering[E]): Trainer[K, V, T] =
-    updateTrees {
-      case (tree, treeIndex, byLeaf) =>
-        val byLeafIndex = byLeaf.map {
-          case (l, instances) =>
-            l.index -> implicitly[Monoid[T]].sum(instances.map { _.target })
-        }
-        tree.prune(byLeafIndex, voter, error)
-    }
-
-  def validate[P, E](error: Error[T, P, E])(implicit voter: Voter[T, P]): Option[E] = {
-    val errors = trainingData.flatMap { instance =>
-      val useTrees = trees.zipWithIndex.filter {
-        case (tree, i) =>
-          sampler.includeInValidationSet(instance.id, instance.timestamp, i)
-      }.map { _._1 }
-      if(useTrees.isEmpty)
-        None
-      else {
-        val prediction = voter.predict(useTrees, instance.features)
-        Some(error.create(instance.target, prediction))
+  def outputStep[O](step: OutputStep[K,V,T,O]): Option[O] = {
+    val treeMap = trees.zipWithIndex.map{case (t,i) => i->t}.toMap
+    var output: Option[O] = None
+    trainingData.foreach{instance =>
+      step.prepare(treeMap, instance).foreach{o =>
+        output = Some(output.map{old => step.semigroup.plus(old, o)}.getOrElse(o))
       }
     }
-    error.semigroup.sumOption(errors)
-  }*/
+    output
+  }
+
+  def updateTargets =
+    trainingStep(UpdateTargets(sampler))
+
+  def expand(implicit splitter: Splitter[V, T], evaluator: Evaluator[V, T], stopper: Stopper[T]) =
+    trainingStep(Expand(sampler, stopper, splitter, evaluator))
+
+  def validate[P, E](error: Error[T, P, E])(implicit voter: Voter[T, P]): Option[E] =
+    outputStep[E](Validate(sampler, error, voter))
 }
 
 object Trainer {
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
@@ -3,56 +3,139 @@ package com.stripe.brushfire.local
 import com.stripe.brushfire._
 import com.twitter.algebird._
 
+
+trait TrainingStep[K,V,T] {
+
+  type K1
+  type V1
+  type V2
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]): Seq[((Int,Int,K1), V1)]
+  def lift(tree: Tree[K,V,T], leafIndex: Int, key: K1, v1: V1): Traversable[V2]
+  def update(tree: Tree[K,V,T], map: Map[Int,V2]): Tree[K,V,T]
+
+  def semigroup1: Semigroup[V1]
+  def semigroup2: Semigroup[V2]
+  def ordering: Ordering[K1]
+}
+
+case class UpdateTargets[K,V,T](sampler: Sampler[K])(implicit val semigroup1: Semigroup[T])
+  extends TrainingStep[K,V,T] {
+
+  type K1 = Unit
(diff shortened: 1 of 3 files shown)


brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
-        fn(tree, index, byLeaf)
-    }
-    copy(trees = newTrees.toList)
+trait TrainingStep[K,V,T] {
+
+  type K1
+  type V1
+  type V2
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]): Seq[((Int,Int,K1), V1)]
```
tixxit: We should make a `case class LeafId(forestIndex: Int, leafIndex: Int)` or something - would probably be generally useful.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+  type V1 = T
+  type V2 = T
+
+  val semigroup2 = semigroup1
+  val ordering = implicitly[Ordering[Unit]]
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]) = {
+     for (
+        (treeIndex, tree) <- trees.toList;
+        i <- 1.to(sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex)).toList;
```
tixxit: Would this be equivalent (I want to avoid calling `leafIndexFor` very often):

```scala
for {
  (treeIndex, tree) <- trees.toList
  leafIndex <- tree.leafIndexFor(instance.features).toList
  _ <- List.fill(sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex))(leafIndex)
} yield (treeIndex, leafIndex, ()) -> instance.target
```

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+
+case class Expand[K,V,T](sampler: Sampler[K], stopper: Stopper[T], splitter: Splitter[V, T], evaluator: Evaluator[V, T])(implicit val ordering: Ordering[K])
+  extends TrainingStep[K,V,T] {
+
+  type K1 = K
+  type V1 = splitter.S
+  type V2 = (K, Split[V, T], Double)
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]) = {
+    lazy val features = instance.features.mapValues { value => splitter.create(value, instance.target) }
```
tixxit: Minor nit - `mapValues` here will recompute the values each time they are accessed, which in our case is O(# trees + # times in training sets). It sucks, but would probably be best to use `map { case (key, value) => (key, splitter.create(value, instance.target)) }` :\

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+    for
+      ((treeIndex, tree) <- trees.toList;
+      i <- 1.to(sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex)).toList;
+      leaf <- tree.leafFor(instance.features).toList if stopper.shouldSplit(leaf.target);
+      (feature, stats) <- features if (sampler.includeFeature(feature, treeIndex, leaf.index)))
+        yield (treeIndex, leaf.index, feature) -> stats
+  }
+
+  def lift(tree: Tree[K,V,T], leafIndex: Int, key: K, v1: V1) = {
+    tree.leafAt(leafIndex).toList.flatMap { leaf =>
```
tixxit: Your inverted tree idea would let us get rid of this full-tree traversal, right?

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+  type V1 = splitter.S
+  type V2 = (K, Split[V, T], Double)
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]) = {
+    lazy val features = instance.features.mapValues { value => splitter.create(value, instance.target) }
+
+    for
+      ((treeIndex, tree) <- trees.toList;
+      i <- 1.to(sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex)).toList;
+      leaf <- tree.leafFor(instance.features).toList if stopper.shouldSplit(leaf.target);
```
tixxit: As above, we could invert this order to avoid recomputing `leafFor`.

tixxit:
I really like the idea overall. Hard to tell if it is too overfit, but I agree that we need to start somewhere!

I'm sort of giving some nit-picky comments to start, but will hopefully give some more useful ones as I understand the abstraction better.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+case class Trainer[K: Ordering, V, T: Monoid](
+    trainingData: Iterable[Instance[K, V, T]],
+    sampler: Sampler[K],
+    trees: List[Tree[K, V, T]]) {
+
+  def trainingStep(step: TrainingStep[K,V,T]): Trainer[K,V,T] = {
+    val treeMap = trees.zipWithIndex.map{case (t,i) => i->t}.toMap
+    var sums1 = Map[(Int,Int,step.K1),step.V1]()
+
+    trainingData.foreach{instance =>
```
tixxit: A bit of golf, but I think this should do roughly the same thing (including laziness):

```scala
val sums1 = MapAlgebra.rollupSum(trainingData.iterator.flatMap(step.prepare(treeMap, _)))
```

tixxit: Wait - I think I misunderstood `rollupSum`. Something similar in spirit should exist though :\

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
-              Nil
-            }
-          }.groupBy { _._2 }
-            .mapValues { _.map { _._1 } }
-        fn(tree, index, byLeaf)
-    }
-    copy(trees = newTrees.toList)
+trait TrainingStep[K,V,T] {
+
+  type K1
```
tixxit: (Just sort of trying to see if I understand these types)

So, this is some key that can be used to segment the initial per-leaf values? I think this could technically be replaced by using a map in `V1`, so is the main reason for this to optimize our m/r steps (since it is reasonable to expect `|K1| = O(|features|)`)?

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+  type V1
```
tixxit: This is a per-leaf (per K1) value that is generated by prepare, based on the instance. We then sum up each value per leaf (per key). This final value is the one given to `lift`.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+  type V2
```
tixxit: This is another per-leaf value, but it obliterates the segmentation by key. So, the first step is to generate a bunch of separate possible values V1, segmented by K1, then we use `semigroup2` + `lift` to basically convert these separate parts into 1 final `V2` value per leaf.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+
+  def prepare(trees: Map[Int, Tree[K,V,T]], instance: Instance[K,V,T]): Seq[((Int,Int,K1), V1)]
+  def lift(tree: Tree[K,V,T], leafIndex: Int, key: K1, v1: V1): Traversable[V2]
+  def update(tree: Tree[K,V,T], map: Map[Int,V2]): Tree[K,V,T]
```
tixxit: One thing that is sort of leaky here, is that the reason we give the entire leaf -> V2 map here is for performance reasons only, right? The alternative being a lot more calls to the slow `leafAt` method. Is this correct?

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -3,81 +3,176 @@ package com.stripe.brushfire.local
...
+
+  def semigroup1: Semigroup[V1]
+  def semigroup2: Semigroup[V2]
+  def ordering: Ordering[K1]
```
tixxit: Is this used by the local trainer?

tixxit:
It seems like the general process for these is to do a series of reductions per-leaf, at least for training. Here's an idea, which I'm a little unsatisfied with, but seems to be mostly equivalent (though a little less opportunity for optimization of the initial prepare key): https://gist.github.com/tixxit/987a9a3130b2b3fa914a

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Heh, so the problem is that I didn't want to call `leafIndexFor` at all in the case where `timesInTrainingSet` returns `0` (which is common). So we need some better optimization than either.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Good to know!

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Same issue as above, but yep.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Yep!

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: It's not. This is a bit of a leak of the needs of Scalding and (presumably) similar platforms.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Hmmm. Well, two things:

- Absent the inverted tree representation, growing a tree is kinda expensive, and so it's nicer to do it all at once. (With the inverted tree representation it would be fine to do it one by one).
- For pruning, rather than expanding/annotating the leaves, you actually do need to do it all at once for the whole tree, so if we had a leaf by leaf update method here we couldn't implement prune.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Yes, exactly.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Right.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
avibryant: Exactly. This is the part that is maybe most gratuitously overfit on our expand step, where `K1` is feature, `V1` is the splitter's joint distribution, and `V2` are candidate splits with a semigroup that picks the single best one.

But basically this is: we want to allow you to start out by computing many things per leaf, then progressively reduce this down to one thing per leaf.

avibryant:
BTW one thing I'm kinda grumpy about is the distinction between `TrainingStep` and `OutputStep`. I wanted each step to be able to either produce new trees or some sidechannel output or both (for example, in the long run I'd really like to compute out-of-band error during the expand step). But `ValidationStep` really does look structurally quite different; it doesn't care about per-leaf or even per-tree, and is instead just computing a single value across the whole forest. A notional `FeatureImportanceStep` would be similar. So I'm not sure how best to model that possibility.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
tixxit: So, I was thinking about this case, and I think it may make more sense to have our general abstraction be something like,

```scala
trait GenTrainer[K, V, T] {
  type TrainingStepContext[K1, V1, V2]

  def trainingStep(step: TrainingStep[K,V,T])(implicit ctx: TrainingStepContext[step.K1, step.V1, step.V2]): Trainer[K,V,T] = ???
}
```

(Or something similar.) In the local case this would basically be a no-op, but for Scalding we may want to capture and `Ordering` and for Spark maybe something else (`ClassTag`s? ¯\_(ツ)_/¯)

tixxit: This of course means we can't *really* treat our training steps too generically, since scalac will need to know the existential types :\

avibryant:
This is very WIP still, but I've gone forward with the `brushfire-training` reorg, because having `TrainingStep` gives me somewhere to land `prune` and `expandInMemory` outside the `Tree` which is still reusable, which was previously a blocker for that. This probably ends up as an overly large PR, but oh well.

PR #78, opened 2015-12-28
WIP: rethink Evaluator

avibryant:
This makes a few related changes to `Evaluator`:
- it decouples it from `Split` and just has it directly operate on `Iterable[T]`, which was the only part of a `Split` it ever cared about anyway.
- this makes it possible to apply an `Evaluator` to a tree or sub-tree, which could be useful in various ways in the future (eg if we want to change `Splitter` to produce candidate sub-trees instead of splits)
- in this context the `Iterable[T]` represents the leaves of the sub-tree, and so it's now labeled as such
- rather than using `Infinity` to represent an unacceptable level of error, it now returns `Option[Double]` so you can return `None` to signal that
- the sign has been reversed (ie this is now an error value not a "goodness" value)
- it doesn't return a new `Split`, just the error
- the root `T` is now also provided. None of the evaluators makes use of it, but there are at least two potential uses I have in mind:
    - considering a split unacceptable based on some ratio of the leaf `T` to the root `T` (eg, "all leaves must contain at least X% of the input Ys")
    - using more global optimization criteria (this comes up in particular in the "explanation tree" use case which I won't explain in more detail here)

This is WIP because I haven't fully updated the training code to make use of it.

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
@@ -46,10 +46,9 @@ trait Split[V, T] {
   def predicates: Iterable[(Predicate[V], T)]
 }
 
-/** Evaluates the goodness of a candidate split */
-trait Evaluator[V, T] {
-  /** returns a (possibly transformed) version of the input split, and a numeric goodness score */
-  def evaluate(split: Split[V, T]): (Split[V, T], Double)
+trait Evaluator[T] {
+  /** returns an overall numeric training error for a tree or split, or None for infinite/unacceptable error */
+  def trainingError(root: T, leaves: Iterable[T]): Option[Double]
 }
 
 /** Provides stopping conditions which guide when splits will be attempted */
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Defaults.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Defaults.scala
@@ -12,7 +12,7 @@ trait LowPriorityDefaults {
 }
 
 trait Defaults extends LowPriorityDefaults {
-  implicit def chiSquaredEvaluator[V, L, W](implicit weightMonoid: Monoid[W], weightDouble: W => Double): Evaluator[V, Map[L, W]] = ChiSquaredEvaluator[V, L, W]
+  implicit def chiSquaredEvaluator[L, W](implicit weightMonoid: Monoid[W], weightDouble: W => Double): Evaluator[Map[L, W]] = ChiSquaredEvaluator[L, W]
   implicit def frequencyStopper[L]: Stopper[Map[L, Long]] = FrequencyStopper(10000, 10)
 
   implicit def intSplitter[T: Monoid]: Splitter[Int, T] = BinarySplitter[Int, T](LessThan(_))
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Evaluators.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Evaluators.scala
@@ -2,16 +2,16 @@ package com.stripe.brushfire
 
 import com.twitter.algebird._
 
-case class ChiSquaredEvaluator[V, L, W](implicit weightMonoid: Monoid[W], weightDouble: W => Double)
-    extends Evaluator[V, Map[L, W]] {
-  def evaluate(split: Split[V, Map[L, W]]) = {
-    val rows = split.predicates.map { _._2 }.filter { _.nonEmpty }
+case class ChiSquaredEvaluator[L, W](implicit weightMonoid: Monoid[W], weightDouble: W => Double)
+    extends Evaluator[Map[L, W]] {
+  def trainingError(root: Map[L, W], leaves: Iterable[Map[L,W]]) = {
+    val rows = leaves.filter { _.nonEmpty }
     if (rows.size > 1) {
       val n = weightMonoid.sum(rows.flatMap { _.values })
       val rowTotals = rows.map { row => weightMonoid.sum(row.values) }.toList
       val columnKeys = rows.flatMap { _.keys }.toList
       val columnTotals = columnKeys.map { column => column -> weightMonoid.sum(rows.flatMap { _.get(column) }) }.toMap
-      val testStatistic = (for {
+      Some((for {
         column <- columnKeys
         (row, index) <- rows.zipWithIndex
       } yield {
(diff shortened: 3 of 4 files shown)


avibryant:
@tixxit ok I think this is ready now for review

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -46,10 +46,9 @@ trait Split[V, T] {
   def predicates: Iterable[(Predicate[V], T)]
 }
 
-/** Evaluates the goodness of a candidate split */
-trait Evaluator[V, T] {
-  /** returns a (possibly transformed) version of the input split, and a numeric goodness score */
-  def evaluate(split: Split[V, T]): (Split[V, T], Double)
+trait Evaluator[T] {
+  /** returns an overall numeric training error for a tree or split, or None for infinite/unacceptable error */
+  def trainingError(root: T, leaves: Iterable[T]): Option[Double]
```
avibryant: It seems a bit arbitrary to make this be a `Double` - it could be some generic `F:Ordering`, but then again I've got a bit of type param exhaustion.

avibryant: I also wonder whether the `Option[T]` is actually confounding two concepts: "unacceptable" should really be something that `Stopper` determines separately. That is, `Evaluator` would always return `Double` but `Stopper` would include both  `isWorthTryingToSplit` and `isAcceptableSplitOutcome`.

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
```diff
@@ -27,6 +27,22 @@ object LeafNode {
 }
 
 case class AnnotatedTree[K, V, T, A: Semigroup](root: Node[K, V, T, A]) {
+
+  private var sumTargetsCache: Option[T] = None
```
avibryant: Not loving this but it can't just be lazy because it needs the monoid. Maybe we should lift the `Monoid[T]` to be something the tree carries around?

example/iris
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-java -Xmx2G -cp ../brushfire-scalding/target/scala-2.10/brushfire-scalding-0.6.0-SNAPSHOT-jar-with-dependencies.jar \
+java -Xmx2G -cp ../brushfire-scalding/target/scala-2.10/brushfire-scalding-0.6.4-SNAPSHOT-jar-with-dependencies.jar \
```
avibryant: This doesn't actually work for me on either my laptop or runway for some reason, but I think unrelated to this PR.

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
tixxit: Yeah, there are sort of 2 issues. One is that caching it is a bit funky, since, technically, we could run `sumTargets` with a different monoid (eg max vs + vs * vs whatever) and so things would subtly break. OTOH, full-tree traversals are generally bad things that we want to avoid. It may be that we do a full tree traversal, but move the caching somewhere else for `sumTargets`?

With annotations, we have the expectation that they are attached to the trees (and both @NathanHowell and @striation just made annotations strict in the inner nodes, which I agree is the right decision, since you can always get laziness with a wrapper type), but `sumTargets` doesn't quite have this (eg it doesn't return a new tree like the annotation methods do).

In any case we should be very explicit about what is going on with some docs on the `sumTargets` method.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
striation: So, I think we should stick with `Double` instead of `Option[Double]`. My guess is that a `NaN` value would be totally fine to use for this purpose (it certainly won't be a valid score).

The idea of using some `Double => Boolean` values to test criteria seems fine to me. I just think we don't need to create more boxes in this case.

(Sorry I am late to this party.)

example/iris
tixxit: Yeah, I should probably just revisit these scripts and fix them all to do something a bit more robust (eg not requiring changes to this file when we bump the version).

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
tixxit: Re: Double vs new type param - I'm definitely pro-`Double` until we reach a point where we have an `Evaluator` where `Double` won't work anymore.

avibryant:
Got rid of the `root` part, and merged with #77 which should now be landed first.

brushfire-training/src/main/scala/com/stripe/brushfire/training/Brushfire.scala
```diff
@@ -54,19 +55,9 @@ case class Split[V, T](predicate: Predicate[V], leftDistribution: T, rightDistri
...
-   *
-   * This method may transform the split, in which case the score
-   * applies to the split that is returned. The result is optional to
-   * handle cases where a split is not valid (for example, if one side
-   * of a split is "empty" in some sense).
-   */
-  def evaluate(split: Split[V, T]): Option[(Split[V, T], Double)]
+trait Evaluator[T] {
+  /** returns an overall numeric training error for a tree or split, or None for infinite/unacceptable error */
+  def trainingError(leaves: Iterable[T]): Option[Double]
```
tixxit: :+1: on the rename!

tixxit: (Also, 1 less type param!)

brushfire-training/src/main/scala/com/stripe/brushfire/training/Evaluators.scala
```diff
@@ -0,0 +1,49 @@
...
+import com.stripe.brushfire._
+import com.twitter.algebird._
+
+case class ChiSquaredEvaluator[L, W](implicit weightMonoid: Monoid[W], weightDouble: W => Double)
+    extends Evaluator[Map[L, W]] {
+  def trainingError(leaves: Iterable[Map[L,W]]) = {
+    val rows = leaves.filter { _.nonEmpty }
+    if (rows.size > 1) {
+      val n = weightMonoid.sum(rows.flatMap { _.values })
+      val rowTotals = rows.map { row => weightMonoid.sum(row.values) }.toList
```
tixxit: I think if we can just switch the order here to:

```scala
val rowTotals = rows.map { row => weightMonoid.sum(row.values) }.toList
val n = weightMonoid.sum(rowTotals)
```

brushfire-training/src/main/scala/com/stripe/brushfire/training/Evaluators.scala
```diff
@@ -0,0 +1,49 @@
...
+      val columnKeys = rows.flatMap { _.keys }.toList
```
tixxit: We probably want `.toSet` here instead of `.toList`.

brushfire-training/src/main/scala/com/stripe/brushfire/training/Evaluators.scala
```diff
@@ -0,0 +1,49 @@
...
+      val columnTotals = columnKeys.map { column => column -> weightMonoid.sum(rows.flatMap { _.get(column) }) }.toMap
+      Some((for {
+        column <- columnKeys
+        (row, index) <- rows.zipWithIndex
+      } yield {
+        val observed = row.getOrElse(column, weightMonoid.zero)
+        val expected = (columnTotals(column) * rowTotals(index)) / n
```
tixxit: Since we're randomly accessing `rowTotals`, we'll probably want to use a `Vector` above, instead of a `List`.

brushfire-training/src/main/scala/com/stripe/brushfire/training/Evaluators.scala
tixxit: Wait - maybe list is what we want? I see we're using it in our for-comprehension below! Will look again...

tixxit: Hrmm, no, I think we do want a set, since we get our per-row iteration with the rows.zipWithIndex.

brushfire-training/src/main/scala/com/stripe/brushfire/training/Evaluators.scala
```diff
@@ -0,0 +1,49 @@
...
+
+import com.stripe.brushfire._
+import com.twitter.algebird._
+
+case class ChiSquaredEvaluator[L, W](implicit weightMonoid: Monoid[W], weightDouble: W => Double)
+    extends Evaluator[Map[L, W]] {
+  def trainingError(leaves: Iterable[Map[L,W]]) = {
+    val rows = leaves.filter { _.nonEmpty }
+    if (rows.size > 1) {
+      val n = weightMonoid.sum(rows.flatMap { _.values })
```
tixxit: We should eagerly convert this to a `Double` here, with `weightDouble`.

brushfire-training/src/main/scala/com/stripe/brushfire/training/Evaluators.scala
```diff
@@ -0,0 +1,49 @@
...
+      }).sum * -1)
+    } else
+      None
+  }
+}
+
+case class MinWeightEvaluator[L, W: Monoid](minWeight: W => Boolean, wrapped: Evaluator[Map[L, W]])
+    extends Evaluator[Map[L, W]] {
+  def trainingError(leaves: Iterable[Map[L,W]]) = {
+    if (leaves.forall {freq => minWeight(Monoid.sum(freq.values))})
```
tixxit: nit: Mind if we move `leaves.forall {freq => minWeight(Monoid.sum(freq.values))}` out to its own `val`?

brushfire-training/src/main/scala/com/stripe/brushfire/local/Lines.scala
```diff
@@ -0,0 +1,63 @@
...
+  override def toString(): String =
+    s"Lines(<over $file with $charset>)"
+}
+
+object Lines {
+  def apply(f: File, cs: String = "UTF-8"): Lines[String] =
+    new Lines[String](f, cs) {
+      def iterator: Iterator[String] =
+        new Iterator[String] {
+          println("Opening " + f)
```
striation: I had this in for debugging but we might want to remove it now.

striation:
We should probably remove all the commented code in this PR once we agree it is really gone.

PR #79, opened 2016-01-14
Support multiple tree implementations with Bonsai

striation:
This is a relatively ambitious pull request. The basic idea is to allow Brushfire to work with many different tree representations, instead of requiring users to use its `Tree` and `AnnotatedTree` types. This flexibility is provided by Bonsai's `TreeOps` (and `FullBinaryTreeOps`) type classes, which provide methods to work with trees in a generic way.

One concrete benefit of this is that users can use the compressed trees provided by Bonsai instead of Brushfire's trees. This also means it should be easier to use Brushfire to train models used in other systems (although I haven't actually done this so I don't have a specific example).

Benchmarking I've done shows that there is no major performance difference here -- some of the optimizations in the PR offset the costs of a more generic model. I think there are probably places we could improve further, but for now this seems reasonable.

I should acknowledge that I still haven't fully answered @NathanHowell's concerns from #76. I would be happy to try the changes @avibryant's suggested if everyone agrees that this is the right way to go.

This PR includes and supercedes #76. I could also rebase this PR against #76 if we would prefer to merge that first.


brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
```diff
@@ -1,25 +1,50 @@
...
+    this match {
+      case SplitNode(k, p, lc0, rc0, a) =>
+        val (n1, lc) = lc0.renumber(nextId)
+        val (n2, rc) = rc0.renumber(n1)
+        (n2, SplitNode(k, p, lc, rc, a))
+      case LeafNode(_, target, annotation) =>
+        (nextId + 1, LeafNode(nextId, target, annotation))
+    }
+
+  def fold[B](f: (Node[K, V, T, A], Node[K, V, T, A], (K, Predicate[V], A)) => B, g: Tuple3[Int, T, A] => B): B =
```
tixxit: Any reason to prefer `Tuple3[Int, T, A] => B` over `((Int, T, A)) => B`?

striation: I thought it was easier to read -- I think it's easy to lose track of which parts are argument lists and which parts are tuples otherwise.

striation: I could change it here to use an argument list, but it's going to be provided as a tuple by the ops type class anyway. But if you'd prefer I can definitely change it.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -42,14 +42,23 @@ trait Splitter[V, T] {
...
+  def createSplitNode[K](feature: K): SplitNode[K, V, T, Unit] =
+    SplitNode(feature, predicate, LeafNode(0, leftDistribution), LeafNode(1, rightDistribution))
 }
 
+
 /** Evaluates the goodness of a candidate split */
 trait Evaluator[V, T] {
   /** returns a (possibly transformed) version of the input split, and a numeric goodness score */
-  def evaluate(split: Split[V, T]): (Split[V, T], Double)
+  def evaluate(split: Split[V, T]): Option[(Split[V, T], Double)]
```
tixxit: May be a question for @avibryant too, but what does `None` mean and could we put that in the docs?

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -1,14 +1,14 @@
 package com.stripe.brushfire
 
+import com.stripe.bonsai.FullBinaryTreeOps
 import com.twitter.algebird._
 
-// type Tree[K, V, T] = AnnotatedTree[K, V, T, Unit]
```
tixxit: Any reason to get rid of this type alias? I think we should either keep it, or rename `AnnotatedTree` to `Tree`, as in @NathanHowell's PR.

striation: I can add it back in -- it didn't seem like it was being used (I thought it was originally commented out but maybe I was wrong). I don't feel strongly about this.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
striation: I'll try to document this, and @avibryant can review it for accuracy.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
```diff
@@ -93,78 +179,59 @@ object TreeTraversal {
...
+    // depending on what our predicate says.
+    lazy val branchF: (Node, Node, BranchLabel[K, V, A]) => Stream[LeafLabel[T, A]] =
+      { case (lc, rc, (k, p, _)) =>
+        p(row.get(k)) match {
+          case Some(true) =>
+            recurse(lc)
+          case Some(false) =>
+            recurse(rc)
+          case None =>
+            val (c1, c2) = r(lc, rc, getAnnotation)
```
tixxit: Not sure how many allocations you are trying to save, but I could see `Reorder` instead taking a `f: (Node, Node) => B` and we could have a lazy `recurseSplit: (Node, Node) => Stream[LeafLabel[T, A]] = recurse(_) append recurse(_)`.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
```diff
@@ -48,20 +54,100 @@ trait TreeTraversal[K, V, T, A] {
...
+ * should be used with each traversal.
+ */
+trait Reorder[A] {
+  def setSeed(seed: Option[String]): Reorder[A]
+  def apply[N](n1: N, n2: N, f: N => A): (N, N)
+}
+
+object Reorder {
+
+  // Traverse into the left node first.
```
tixxit: Scaladoc instead?

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
```diff
@@ -15,12 +16,14 @@ import com.twitter.algebird._
  * features). A tree traversal chooses which paths to go down (which may be all
  * of them) and the order in which they are traversed.
  */
-trait TreeTraversal[K, V, T, A] {
+trait TreeTraversal[Tree, K, V, T, A] {
```
tixxit: Maybe rename `Tree` to `TreeRep`? Then we can either add the alias back without much confusion, or just rename `AnnotatedTree` to `Tree`. WDYT?

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
striation: Does this help?

```scala
  /**
   * Evaluate the fitness of a candidate split.
   *
   * This method may transform the split, in which case the score
   * applies to the split that is returned. The result is optional to
   * handle cases where a split is not valid (for example, if one side
   * of a split is "empty" in some sense).
   */
```

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
```diff
@@ -19,37 +19,31 @@ object Tree {
...
-            None
+        val splits = for {
+          (f, s) <- featureMap.toList
+          split <- splitter.split(leaf.target, s)
+          tpl <- evaluator.evaluate(split)
+        } yield (f, tpl)
+
+        if (splits.isEmpty) None else {
+          val (splitFeature, (Split(pred, left, right), _)) = splits.maxBy { case (f, (x, s)) => s }
+          def ex(dist: T): Node[K, V, T, Unit] = {
```
tixxit: What does `ex` mean?

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
striation: Yeah I did have the idea that maybe `Reorder` could just be responsible for the complete traversal. I do think it would be a bit nicer... I had talked myself out of it but you convinced me. I'll code it up and see what happens.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
striation: Yes, I'll do this. Sorry for being so lazy apparently!

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
striation: I think it's just meant to abbreviate "expand" which is what we're doing to the right and left children. It's basically just some pre-processing before calling `expand` recursively.

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
tixxit: No strong preference, more curiosity. It seems a bit weird to use a tuple if you know nothing about bonsai, but we are using it, so I think this is fine.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
tixxit: :+1:

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
tixxit: We definitely use it in our other projects, but maybe not within Brushfire itself.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
tixxit: To be clear, re-order wouldn't be handling the traversal, just the ordering, but instead of returning a `tuple` from `apply`, we'd take 1 more function argument:

```
def apply[B](n1: N, n2: N, a: A)(f: (N, N) => B): B
```

Is that what you were thinking too?

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
johnynek: +1 to easier to read.

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
```diff
@@ -93,84 +122,84 @@ case class AnnotatedTree[K, V, T, A: Semigroup](root: Node[K, V, T, A]) {
...
+    leafAt(leafIndex, root)
 
   /**
-   * Returns the leaf with index `leafIndex` that is a descendant of `start` by
-   * performing a DFS starting with `start`.
+   * Returns the leaf with index `leafIndex` that is a descendant of
+   * `start` by performing a DFS starting with `start`.
    */
-  def leafAt(leafIndex: Int, start: Node[K, V, T, A]): Option[LeafNode[K, V, T, A]] = {
+  def leafAt(leafIndex: Int, start: Node[K, V, T, A]): Option[LeafNode[K, V, T, A]] =
```
avibryant: Should we even have the basic `leafAt` hardcoded to DFS or should we change everything to use traversals?

avibryant:
Just to be clear about the scope here: it doesn't make it possible to train using multiple representations (which is a good thing, I think); it just introduces a generalized traversal abstraction that we can use in other contexts (like evaluation) with other tree representations. Right?

striation:
@avibryant Right, exactly.

brushfire-core/src/main/scala/com/stripe/brushfire/Tree.scala
striation: So it actually looks like type alias still exists in `package.scala`. I'm not sure why that comment is there -- I'm going to remove it.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
striation: Since the alias was never actually removed I think things are probably fine as they are.

brushfire-core/src/main/scala/com/stripe/brushfire/TreeTraversal.scala
striation: Exactly.

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
striation: There's only one use of this method in the trainer -- I'll see if I can replace it with a traversal.

striation: I think currently the traversals we have are somewhat specialized to the tree structure itself, and can't easily accommodate searching for nodes by the leaf ID. For now I'd just as soon leave this in, although I guess it's also possible to make traversals a bit more general.

striation:
@johnynek I'm happy to make the seed an explicit parameter -- someone else will probably seed with the clock, but that's OK (these don't need to be cryptographically secure AFAIK).

striation:
I think the initial goal with `Reorder` was to make the instances more reusable. Since that goal has failed I'll look into "inlining" more of the rest of it in the instance.

striation:
So to answer @johnynek's question in a long-winded sort of way:

It's difficult to know the `N` type at the point we are currently creating `Reorder` instances. This is because of the the way we use path-dependent types to specify the node type of a tree. The `f` is just collateral damage.

I'll just comment this for now. It's possible that we could fix the design (and if we're unhappy enough with the current one I'll take a crack) but it would likely require some changes to *bonsai* to happen.

striation:
@johnynek What do you think? Seems reasonable?

johnynek:
:+1:

tixxit:
Awesome. I think the only thing I think we should do is bump the minor version at least, so that we can indicate a fairly big change (and also possibly release patches to the previous version should migration prove difficult, though I don't suspect we'll have too much trouble).

@NathanHowell Any concerns with this plan?

striation:
It seems like everyone has signed-off on this, and given that we do have a plan to deal with multi-way splits in the future, I'm going to merge this. Thanks everyone!

PR #80, opened 2016-01-27
[WIP] Feature encoders

tixxit:
This is super early work, but we have a CsvTrainerJob, which can run on ~arbitrary CSVs, with the labels provided by the user. The actual types of the values will be inferred before training.

### DIFF
--- a/brushfire-features/src/main/scala/com/stripe/brushfire/features/CsvRowFeatureMapping.scala
+++ b/brushfire-features/src/main/scala/com/stripe/brushfire/features/CsvRowFeatureMapping.scala
@@ -1,0 +1,87 @@
+package com.stripe.brushfire
+package features
+
+import scala.util.Try
+
+import com.twitter.algebird.{ Aggregator, Semigroup, MapMonoid }
+
+object CsvRowFeatureMapping {
+  def isDouble(s: String): Boolean = Try(s.toDouble).isSuccess
+
+  case class NominalCount(uniques: Option[Set[String]]) {
+    def ++ (that: NominalCount): NominalCount =
+      NominalCount(for {
+        a <- this.uniques
+        b <- that.uniques
+        c = a ++ b
+        if (c.size < 20)
+      } yield c)
+  }
+
+  case class Counts(doubles: Long, nominals: NominalCount, count: Long) {
+    def ++(that: Counts): Counts =
+      Counts(this.doubles + that.doubles, this.nominals ++ that.nominals, this.count + that.count)
+
+    def featureType: FeatureType = {
+      val isNumeric = doubles == count
+      val isSmall = nominals.uniques.isDefined
+      (isNumeric, isSmall) match {
+        case (true, true) => FeatureType.Ordinal
+        case (true, false) => FeatureType.Continuous
+        case (false, true) => FeatureType.Nominal
+        case (false, false) => FeatureType.Sparse
+      }
+    }
+  }
+
+  object Counts {
+    def apply(value: String): Counts =
+      Counts(if (isDouble(value)) 1 else 0, NominalCount(Some(Set(value))), 1)
+
+    implicit val CountsSemigroup: Semigroup[Counts] =
+      Semigroup.from[Counts](_ ++ _)
+
+    val aggregator: Aggregator[String, Counts, FeatureType] =
+      new Aggregator[String, Counts, FeatureType] {
+        def prepare(input: String): Counts = Counts(input)
+        def semigroup: Semigroup[Counts] = CountsSemigroup
+        def present(reduction: Counts): FeatureType = reduction.featureType
+      }
+  }
+
+  case class IndexedSeqAggregator[-A, B, +C](agg: Aggregator[A, B, C])
+  extends Aggregator[IndexedSeq[A], IndexedSeq[B], IndexedSeq[C]] {
+    def prepare(input: IndexedSeq[A]): IndexedSeq[B] =
+      input.map { a => agg.prepare(a) }
+    val semigroup: Semigroup[IndexedSeq[B]] =
+      Semigroup.indexedSeqSemigroup(agg.semigroup)
+    def present(reduction: IndexedSeq[B]): IndexedSeq[C] =
+      reduction.map { b => agg.present(b) }
+  }
+
+  def apply(header: CsvRow, featureTypes: IndexedSeq[FeatureType]): FeatureMapping[CsvRow] = {
+    val mapping: Map[String, (FeatureType, CsvRow => FeatureValue)] = {
+      def extractor(index: Int, featureType: FeatureType): CsvRow => FeatureValue = {
+        def mk(f: String => FeatureValue): CsvRow => FeatureValue = row => f(row(index))
+        featureType match {
+          case FeatureType.Ordinal => mk(x => Dispatched.ordinal(x.toDouble))
+          case FeatureType.Nominal => mk(Dispatched.nominal)
+          case FeatureType.Continuous => mk(x => Dispatched.continuous(x.toDouble))
+          case FeatureType.Sparse => mk(Dispatched.sparse)
+        }
+      }
+
+      (header zip featureTypes).zipWithIndex.map { case ((key, ft), i) =>
+        key -> (ft, extractor(i, ft))
+      } (collection.breakOut)
+    }
+    FeatureMapping(mapping)
+  }
+
+  def aggregator(header: CsvRow): Aggregator[CsvRow, IndexedSeq[Counts], FeatureMapping[CsvRow]] =
+    IndexedSeqAggregator(Counts.aggregator)
+      .andThenPresent(apply(header, _))
+
+  def apply(header: CsvRow, rows: TraversableOnce[CsvRow]): FeatureMapping[CsvRow] =
+    aggregator(header).apply(rows)
+}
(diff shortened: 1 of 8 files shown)


brushfire-features/src/main/scala/com/stripe/brushfire/features/CsvRowFeatureMapping.scala
```diff
@@ -0,0 +1,87 @@
...
+object CsvRowFeatureMapping {
+  def isDouble(s: String): Boolean = Try(s.toDouble).isSuccess
+
+  case class NominalCount(uniques: Option[Set[String]]) {
+    def ++ (that: NominalCount): NominalCount =
+      NominalCount(for {
+        a <- this.uniques
+        b <- that.uniques
+        c = a ++ b
+        if (c.size < 20)
```
avibryant: 20 seems pretty small, I'd think we'd want to allow at least 100 or 200 here.

brushfire-features/src/main/scala/com/stripe/brushfire/features/CsvRowFeatureMapping.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    val aggregator: Aggregator[String, Counts, FeatureType] =
+      new Aggregator[String, Counts, FeatureType] {
+        def prepare(input: String): Counts = Counts(input)
+        def semigroup: Semigroup[Counts] = CountsSemigroup
+        def present(reduction: Counts): FeatureType = reduction.featureType
+      }
+  }
+
+  case class IndexedSeqAggregator[-A, B, +C](agg: Aggregator[A, B, C])
```
avibryant: We should pull this into algebird, obv.

brushfire-features/src/main/scala/com/stripe/brushfire/features/FeatureEncoder.scala
```diff
@@ -0,0 +1,27 @@
...
+sealed trait FeatureType
+object FeatureType {
+  case object Ordinal extends FeatureType
+  case object Nominal extends FeatureType
+  case object Continuous extends FeatureType
+  case object Sparse extends FeatureType
+}
+
+case class FeatureMapping[A](
+  mapping: Map[String, (FeatureType, A => FeatureValue)]
```
avibryant: Why do we even need `FeatureType`? And if we don't need it, I think this would be much better as `(Map[String,A] => Map[String,FeatureValue])`.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/CsvTrainer.scala
```diff
@@ -0,0 +1,74 @@
...
+      .map(splitRow)
+
+  def guessFeatureMapping(header: CsvRow, rows: TypedPipe[CsvRow]): Execution[FeatureMapping[CsvRow]] =
+    rows
+      .aggregate(CsvRowFeatureMapping.aggregator(header))
+      .toIterableExecution
+      .map(_.head)
+
+  val trainingData: Execution[TypedPipe[Instance[String, FeatureValue, Map[String, Long]]]] =
+    guessFeatureMapping(header, rows).map { mapping =>
```
avibryant: we should build this guessFeatureMapping step into `Trainer`...

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/CsvTrainer.scala
```diff
@@ -0,0 +1,74 @@
...
+  import JsonInjections._
+
+  implicit val stopper = FrequencyStopper[String](10, 3)
+  implicit val errorOrdering = Ordering[Double].on[AveragedValue] { av: AveragedValue => av.value } // Allows us to compare BrierScoreErrors.
+
+  val error = BrierScoreError[String, Long]
+
+  def splitRow(row: String): CsvRow = row.split(",").map(_.trim)
+
+  val header: CsvRow = splitRow(args("features"))
```
avibryant: this seems to be `columns` rather than `features` since it also includes the label feature

brushfire-features/src/main/scala/com/stripe/brushfire/features/FeatureEncoder.scala
tixxit: `A` here is actually the *record* type, not a value. So `A` is a `Map[String, String]` in the CSV case or some `A <: ThriftStruct` in the thrift case. That said, this could be `A => Map[String, FeatureValue]`. I didn't do this for some reasons initially, but that reason may no longer be valid :D Let me just try to remember my thought process.

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/CsvTrainer.scala
tixxit: agreed!

tixxit:
@avibryant So, this has some of the stuff I've been doing (still very WIP), but the important bits are:

 * `FValue` ADT, which is basically just `Number | Text | Boolean` right now
 * `FeatureParser[-A]` that is basically just `A => Map[String, FValue]` (eg `A` could be `CsvRow`)
   * there is also a `TrainingDataFeatureParser[-A]` which includes ID/timestamp/target parsers
 * `FeatureEncoder[K, +V]` that can convert some `Map[String, FValue] => Map[K, V]`
 * `FeatureEncoding[K, V, T]` that holds a `FeatureEncoder[K, V]` and `Splitter[V, T]` (we can get rid of the `T` after your PR lands)
 * `TrainingPlatform#Trainer[-A, +B]` which describes some series of passes over data of type `A` to end up with a `B`

There is an implementation, `DispatchedFeatureEncoding`, of a `FeatureEncoding` for the dispatched type, which has a "trainer" that does a pass over the data and attempts to infer the sub-type of Dispatched to use for it.

The main goal of the `FeatureParser` vs `FeatureEncoder` split is so that we can separate the input type from the input-type agnostic feature encoding bits from the tree K/V type. So, we can train of CSV data or thrift and still write a web service that accepts JSON.

brushfire-features/src/main/scala/com/stripe/brushfire/features/DispatchedFeatureEncoder.scala
```diff
@@ -0,0 +1,232 @@
...
+
+  def isDouble(s: String): Boolean = Try(s.toDouble).isSuccess
+
+  case class UniqueCount[A](uniques: Option[Set[A]]) {
+    def ++ (that: UniqueCount[A]): UniqueCount[A] =
+      UniqueCount(for {
+        a <- this.uniques
+        b <- that.uniques
+        c = a ++ b
+        if (c.size < 20)
```
tixxit: As @avibryant mentioned earlier, this should probably be bumped up or made configurable.

brushfire-features/src/main/scala/com/stripe/brushfire/features/DispatchedFeatureEncoder.scala
```diff
@@ -0,0 +1,232 @@
+package com.stripe.brushfire
+package features
+
+import scala.util.Try
+
+import com.twitter.algebird.{ Aggregator, Semigroup, Monoid }
+
+case class DispatchedFeatureEncoding[L](encoder: DispatchedFeatureEncoder)
+extends FeatureEncoding[String, FeatureValue, Map[L, Long]] with Defaults {
```
tixxit: Sorry for the terrible name collisions here, but `FeatureValue` is a type alias for `Dispatched[Double, String, Double, String]`

brushfire-features/src/main/scala/com/stripe/brushfire/features/DispatchedFeatureEncoder.scala
```diff
@@ -0,0 +1,232 @@
...
+    )
+}
+
+object DispatchedFeatureEncoding {
+  import DispatchedFeatureEncoder.Counts
+
+  def trainer[L](
+    platform: TrainingPlatform
+  )(implicit
+    ev: platform.AggregateContext[Map[String, Option[Counts]]]
```
avibryant: I need to read more of this code but it seems like this and `TrainingStep` might be coming at the same problem from two different angles and we should reconcile them?

brushfire-features/src/main/scala/com/stripe/brushfire/features/FeatureParser.scala
```diff
@@ -0,0 +1,59 @@
...
+    platform.Trainer.pure(CsvRowFeatureParser(header))
+}
+
+case class CsvRowTrainingDataParser(
+  header: CsvRow,
+  timestamp: Option[String],
+  id: Seq[String],
+  target: String
+) extends TrainingDataParser[CsvRow, Map[String, Long]] {
+  private def indexOf(tpe: String)(col: String): Int = {
```
avibryant: small thing but I found the currying here hard to read; maybe a slightly more verbose use of a single argument list would be better

tixxit: :+1:

brushfire-features/src/main/scala/com/stripe/brushfire/features/DispatchedFeatureEncoder.scala
tixxit: Yeah - I'll take a look again and see if we can unify these things - or what the minimum is needed to build both on top of something else!

PR #82, opened 2016-03-10
Add annotation support to json serializers.

tixxit:
This let's us (de)serialize our annotations, which we previously just didn't support.

cc @avibryant @striation

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
@@ -113,6 +113,11 @@ object JsonInjections {
     implicit kInj: JsonNodeInjection[K],
     pInj: JsonNodeInjection[T],
     vInj: JsonNodeInjection[V],
+    // A common case is that A is unit, so we special case it here and avoid
+    // serializing a whole bunch of units (and finding a sane serialization for
+    // it). We do this by using `WithFallback`. If A is a unit, then we don't
+    // need an injection. OTOH, if A isn't Unit, then we get an injection and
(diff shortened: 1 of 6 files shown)


brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
```diff
@@ -109,28 +109,35 @@ object JsonInjections {
...
-        def apply(node: Node[K, V, T, Unit]) = node match {
-          case LeafNode(index, target, _) =>
+    implicit def nodeJsonNodeInjection: JsonNodeInjection[Node[K, V, T, A]] =
+      new AbstractJsonNodeInjection[Node[K, V, T, A]] {
+        def apply(node: Node[K, V, T, A]) = node match {
+          case LeafNode(index, target, annotation) =>
             val obj = JsonNodeFactory.instance.objectNode
             obj.put("leaf", toJsonNode(index))
             obj.put("distribution", toJsonNode(target))
+            maybeAInj.withFallback { aInj =>
```
striation: I like the use of `WithFallback` to avoid serializing `Unit` but maybe a comment explaining what is going on (either here or in the method parameters) would be helpful.

striation:
Looks reasonable to me. I'm glad you moved the semigroup instance from the tree to the methods that need it. :+1:

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
tixxit: Sounds good!

PR #83, opened 2016-03-11
Model migration script

tixxit:
cc @striation @avibryant @brahn 

This adds a migration script that can upgrade compatible models from the old format to the new format. Compatible here means that all splits are essentially binary splits based on a single feature/predicate. This will let us just upgrade our current models to the new format when we make the switch to brushfire 0.7.0.

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
@@ -36,6 +36,9 @@ case class SplitNode[K, V, T, A](key: K, predicate: Predicate[V], leftChild: Nod
     }
 
   def splitLabel: (K, Predicate[V], A) = (key, predicate, annotation)
+
+  override def toString: String =
+    s"SplitNode($key,${predicate.display(key.toString)},$leftChild,$rightChild)"
 }
 
 object SplitNode {
--- a/scripts/migrate-to-0.7
+++ b/scripts/migrate-to-0.7
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+from __future__ import print_function
+import argparse
+import json
+import sys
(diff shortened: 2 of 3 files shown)


scripts/migrate-to-0.7
```diff
@@ -0,0 +1,57 @@
...
+    """
+    if type(node) == list:
+        if len(node) != 2:
+            raise MigrationFailed('split node is not binary')
+        left, right = node
+        key = left['feature']
+        if right['feature'] != key:
+            raise MigrationFailed('predicates use different feature')
+        predicate = left['predicate']
+        if right['predicate'].get('not') != predicate:
```
striation: If the left side was a `Not(_)` will the right side be a double-Not? I think it is, but if not it would be a problem.

tixxit: Yeah - I was thinking about this. None of the default splitters would do anything smart here, but it's possible someone may have a fancier splitter that does. It certainly doesn't hurt to handle the symmetric case anyways!

striation:
Well -- if you want to handle the not case that would be fine, but if you want to merge this I'd also be ok. :+1:

brushfire-core/src/main/scala/com/stripe/brushfire/AnnotatedTree.scala
```diff
@@ -36,6 +36,9 @@ case class SplitNode[K, V, T, A](key: K, predicate: Predicate[V], leftChild: Nod
     }
 
   def splitLabel: (K, Predicate[V], A) = (key, predicate, annotation)
+
+  override def toString: String =
+    s"SplitNode($key,${predicate.display(key.toString)},$leftChild,$rightChild)"
```
striation: Nice!

striation:
Looks solid to me! :+1:

PR #84, opened 2016-03-14
Make Predicate[V] easier to work with.

striation:
In line with the move toward binary splits, we can simplify our
predicates to make them easier to work with. Previously we've relied a
lot on ad-hoc encodings, (such as `Not(LessThan(_))` and
`OneOf(LessThan(_), EqualTo(_))`).

This commit replaces the previous predicate hierarchy with a set of
six non-recursive predicate types:

 - IsEq
 - NotEq
 - Lt
 - LtEq
 - Gt
 - GtEq

Predicates will display themselves in a more friendly format, which
should make them easier to work with. They also now come with a
negation operator (!p) which should remove the necessity of
creating/matching Not(_) nodes in user code.

The Ordering[V] requirement is moved from the Predicate[V]
constructors into the apply() and run() methods. This will reduce the
memory use for predicates, and also simplifies the serialization code.

The support for IsPresent(_) has also been removed. If you want
missing values to be treated specially (i.e. treated as just matching
or not matching), you'll need to arrange for that to happen before the
row value is constructed. Future work might generalize rows from
Map[K, V] to (K => V) to make this process more efficient.

?r by @tixxit, @avibryant, and/or @jvns.

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
@@ -44,6 +44,9 @@ trait Splitter[V, T] {
 /** Candidate split for a tree node */
 case class Split[V, T](predicate: Predicate[V], leftDistribution: T, rightDistribution: T) {
 
+  def map[U](f: V => U): Split[U, T] =
+    Split(predicate.map(f), leftDistribution, rightDistribution)
+
   /**
    * Given a feature key, create a SplitNode from this Split.
    *
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
@@ -8,12 +8,14 @@ import com.twitter.bijection.Inversion.{ attempt, attemptWhen }
 import com.twitter.bijection.InversionFailure.{ failedAttempt, partialFailure }
 import JsonNodeInjection._
 import org.codehaus.jackson.JsonNode
+import org.codehaus.jackson.map.JsonMappingException
 import org.codehaus.jackson.node.JsonNodeFactory
 import scala.util._
 import scala.collection.JavaConverters._
 import scala.util.control.NonFatal
 
 object JsonInjections {
+
   implicit def mapInjection[L, W](implicit labelInj: Injection[L, String], weightInj: JsonNodeInjection[W]): JsonNodeInjection[Map[L, W]] =
     new AbstractJsonNodeInjection[Map[L, W]] {
       def apply(frequencies: Map[L, W]) =
@@ -28,7 +30,8 @@ object JsonInjections {
     }
 
   implicit def dispatchJsonNodeInjection[A: JsonNodeInjection, B: JsonNodeInjection, C: JsonNodeInjection, D: JsonNodeInjection]: JsonNodeInjection[Dispatched[A, B, C, D]] = new AbstractJsonNodeInjection[Dispatched[A, B, C, D]] {
-    def apply(dispatched: Dispatched[A, B, C, D]) = {
+
+    def apply(dispatched: Dispatched[A, B, C, D]): JsonNode = {
       val obj = JsonNodeFactory.instance.objectNode
       dispatched match {
         case Ordinal(v) => obj.put("ordinal", toJsonNode(v))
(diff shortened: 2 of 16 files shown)


striation:
The premise behind this PR is that in the future if we want to change how missing values affect traversal, we would use feature encoding. (I *think* that's where we ended up during our last conversation.)

avibryant:
> The premise behind this PR is that in the future if we want to change how missing values affect traversal, we would use feature encoding

I think that's basically right, although one clarification/question: rather than being something we do during feature encoding, I think it would need to be something we do during training/splitting. For example, for a bag of words feature, I'd expect the `V` to be `Set[String]`, and this would be expanded into a number of "feature regions" during the training phase which would be aggregated separately. (Incidentally now that I think of this, it seems like a useful optimization for other feature types as well). But this means that the predicates generated would be things like `GtEq(Set("foo"))`, with the meaning "contains the token 'foo'". That would mean we'd need a special Ordering to achieve those semantics, which we'll need to make available to anyone using these predicates. Does that seem reasonable?

jvns:
This looks easier to understand to me!

tixxit:
@avibryant Yeah - but I think in that case we'd go further and actually want our own wrapper around `Set` that guarantees the `Ordering` we expect.

Another option Erik and I were thinking about was that during feature encoding, you'd presumably know of all possible values that can be seen. If we think of a `row` as just a `K => Option[V]`, then we could easily have a "feature encoder" that can provide defaults for all valid values of `K`. That is, if we wanted to have a word-per feature encoding. I do prefer the set style with smarter splits though.

brushfire-core/src/main/scala/com/stripe/brushfire/Brushfire.scala
```diff
@@ -44,6 +44,9 @@ trait Splitter[V, T] {
 /** Candidate split for a tree node */
 case class Split[V, T](predicate: Predicate[V], leftDistribution: T, rightDistribution: T) {
 
+  def map[U](f: V => U): Split[U, T] =
+    Split(predicate.map(f), leftDistribution, rightDistribution)
```
tixxit: I think there is a `mapPredicates` or something function on `AnnotatedTree` that does this manually - may be nice to rewrite it to use this `map` instead.

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
```diff
@@ -73,39 +73,31 @@ object JsonInjections {
...
         }
         obj
       }
 
-      override def invert(n: JsonNode) = {
+      override def invert(n: JsonNode) =
         n.getFieldNames.asScala.toList.headOption match {
-          case Some("eq") => fromJsonNode[V](n.get("eq")).map { EqualTo(_) }
-          case Some("lt") =>
-            if (ord == null)
```
tixxit: Good riddance!

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
```diff
@@ -73,39 +73,31 @@ object JsonInjections {
...
-            else
-              fromJsonNode[V](n.get("lt")).map { LessThan(_) }
-          case Some("not") => fromJsonNode[Predicate[V]](n.get("not")).map { Not(_) }
-          case Some("or") => fromJsonNode[List[Predicate[V]]](n.get("or")).map { AnyOf(_) }
-          case Some("exists") =>
-            val predNode = n.get("exists")
-            if (predNode.isNull) Success(IsPresent[V](None))
-            else fromJsonNode[Predicate[V]](predNode).map(p => IsPresent(Some(p)))
-          case _ => sys.error("Not a predicate node: " + n)
+          case Some("iseq") => fromJsonNode[V](n.get("iseq")).map(IsEq(_))
```
tixxit: Bikeshed moment, but I'd prefer camelcase (or even snakecase) over no case.

brushfire-core/src/main/scala/com/stripe/brushfire/Predicate.scala
```diff
@@ -1,99 +1,115 @@
...
+  import Predicate._
 
-  def run(o: Option[V]): Boolean =
-    apply(o) match {
-      case Some(b) => b
-      case None => true
+  /**
+   * Evaluate this predicate for the feature value `v`.
+   */
+  def apply(v: V)(implicit ord: Ordering[V]): Boolean =
```
tixxit: :+1: on requiring the ordering here.

tixxit:
Nothing to do with this PR, but for things like the `Set` case, we may, technically, want to use a partial ordering instead of an ordering.

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
striation: Sure I can change that.

striation:
Yeah based on @avibryant's comment I think using a partial ordering makes sense. We can do that in another PR.

brushfire-core/src/main/scala/com/stripe/brushfire/Injections.scala
```diff
@@ -39,13 +42,14 @@ object JsonInjections {
...
-      case Some("sparse") => fromJsonNode[D](n.get("sparse")).map { Sparse(_) }
-      case _ => sys.error("Not a dispatched node: " + n)
-    }
+    override def invert(n: JsonNode): Try[Dispatched[A, B, C, D]] =
+      n.getFieldNames.asScala.toList.headOption match {
+        case Some("ordinal") => fromJsonNode[A](n.get("ordinal")).map { Ordinal(_) }
+        case Some("nominal") => fromJsonNode[B](n.get("nominal")).map { Nominal(_) }
+        case Some("continuous") => fromJsonNode[C](n.get("continuous")).map { Continuous(_) }
+        case Some("sparse") => fromJsonNode[D](n.get("sparse")).map { Sparse(_) }
+        case _ => failedAttempt("Not a dispatched node: " + n)
```
tixxit: Thanks for fixing these!

tixxit:
:+1:

PR #85, opened 2016-03-15
This commit tightens up Brushfire's JSON injections.

striation:
Previously, we would accept a lot of values which
weren't precisely what Brushfire produced. In my
opinion this was potentially error-prone, since it
might cause future parsing probles that we woudn't
easily detect.

It also adds a test to ensure that our injections
will never throw exceptions on the inverse function,
even when fed arbitrary JSON.


tixxit:
Verified against a couple of our migrated models :+1:

PR #86, opened 2016-03-17
Require full orderings for nominal/sparse

tixxit:
Now that we *always* use `Ordering[V]` for all predicate comparisons, we need Dispatched's Ordering to actually handle nominal/sparse cases. This isn't the best solution, since it restricts our nominal/sparse types unnecessarily, but is meant as a stopgap until we switch to using PartialOrderings.


striation:
Looks good to me! :+1:

PR #87, opened 2016-03-22
Use partial order to support different split behaviors

erik-stripe:
This PR moves brushfire-core over to use `spire.algebra.PartialOrder` and `spire.algebra.Order` where applicable. At this point we don't actually _use_ any partial orderings, but the support is there.

Before we merge this I'd like to add some tests to prove this is working, and maybe flesh out support for the _is present_ use-case a bit. What do you all think we need?

Review by @tixxit and @avibryant.

### DIFF
--- a/brushfire-core/build.sbt
+++ b/brushfire-core/build.sbt
@@ -11,6 +11,7 @@ libraryDependencies ++= {
     jacksonXC,
     jacksonJAXRS,
     tDigest,
+    spire,
     scalaTest,
     scalaCheck
   )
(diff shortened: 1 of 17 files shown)


tixxit:
Tests are good, but I don't think we need to flesh out the is-present stuff in this PR. Just moving `Dispatched` over to using a `PartialOrder` would be enough.

striation:
OK, that makes sense. I'll do that next.

striation:
Alright, `Dispatched` is moved over (that was easy!).

striation:
I updated this to master, should be able to be merged.

?r @tixxit @johnynek

johnynek:
This looks good to me.

I do think if we are going to use spire more (and we should) we will get mileage out of the algebra work to make algebird, scalding and spire work better together.

That or give up and make a `spire-algebird` package that gives conversions between the various typeclasses.

PR #88, opened 2016-03-25
Erik non lazy initialization

striation:
This is a small change that improves the efficiency of our traversals.

@jvns noticed that accessing the `lazy val` function instances was slowing down our traversals. In this case (since they are local variables only used by one thread) we can replace the `lazy val` variables with `var` variables which we initialize as soon as it is legal (we can't use `val` due to an overly-cautious circularity check done by scalac).

The tests pass and @jvns has verified that this branch seems to improve the performance of the traversals.

?r @tixxit @jvns @avibryant


tixxit:
:+1:

PR #89, opened 2016-03-27
Bounded-memory local training

avibryant:
Some work from the plane, ?r @striation @tixxit 

This has the same goal as #77 but is deliberately worse code (for now) in the interests of being less invasive and getting merged quickly. It completely avoids making any useful abstractions and instead just provides the most direct implementation of a single-node, single-pass-per-expansion local trainer.

The only interesting thing it does is this: by streaming over the training data we avoid using O(training set) memory, but if we try to expand an entire level at once, we still have an O(2^depth * features) problem. So `expand` takes a parameter of the maximum number of tree leaves to try to expand, per tree, in any one pass, and randomly picks which ones to do (using something like reservoir sampling to get a uniform sample of leaves that don't meet the stopping criteria). This lets you trade off memory use vs. performance (by forcing more passes but capping the memory).

There's still the need to keep all of the trees in memory at once, so it's not *truly* constant memory, but that's harder to avoid. If it becomes a problem we can look at using bonsai representations even during training...

### DIFF
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
@@ -31,10 +31,5 @@ object Example extends Defaults {
       println(trainer.validate(AccuracyError()))
       println(trainer.validate(BrierScoreError()))
     }
-
-    implicit val ord = Ordering.by[AveragedValue, Double] { _.value }
-    trainer = trainer.prune(BrierScoreError())
-    println(trainer.validate(AccuracyError()))
-    println(trainer.validate(BrierScoreError()))
   }
 }
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
@@ -9,28 +9,29 @@ import AnnotatedTree.{AnnotatedTreeTraversal, fullBinaryTreeOpsForAnnotatedTree}
 object Example extends Defaults {
 
   def main(args: Array[String]) {
-    val cols = args.toList
-  
-    val trainingData = io.Source.stdin.getLines.map { line =>
+    val path = args.head
+    val cols = args.tail.toList
+
+    val trainingData = Lines(path).map { line =>
       val parts = line.split(",").reverse.toList
       val label = parts.head
       val values = parts.tail.map { s => s.toDouble }
       Instance(line, 0L, Map(cols.zip(values): _*), Map(label -> 1L))
-    }.toList
-  
+    }.toIterable
+
     var trainer =
       Trainer(trainingData, KFoldSampler(4))
         .updateTargets
-  
+
     println(trainer.validate(AccuracyError()))
     println(trainer.validate(BrierScoreError()))
-  
+
     1.to(10).foreach { i =>
       trainer = trainer.expand(1)
       println(trainer.validate(AccuracyError()))
       println(trainer.validate(BrierScoreError()))
     }
-  
+
     implicit val ord = Ordering.by[AveragedValue, Double] { _.value }
     trainer = trainer.prune(BrierScoreError())
     println(trainer.validate(AccuracyError()))
--- a/brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
+++ b/brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
@@ -24,12 +24,10 @@ object Example extends Defaults {
         .updateTargets
 
     println(trainer.validate(AccuracyError()))
-    println(trainer.validate(BrierScoreError()))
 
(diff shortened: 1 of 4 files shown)


brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -6,81 +6,145 @@ import com.twitter.algebird._
...
+
+        r
+      }
+    }
+  }
+
+  def containsKey(key: A): Boolean = randValue(key) <= threshold
+  def update(key: A, value: B) {
+    if(containsKey(key))
+      mapValues += key -> value
```
striation: Since we're just mutating this structure in place why don't we use `mutable.Map` for better performance?

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -6,81 +6,145 @@ import com.twitter.algebird._
...
+      ) {
+        var leafStats = treeStats.get(leafIndex).getOrElse(Map[K,splitter.S]())
+        val combined = leafStats.get(feature) match {
+          case Some(old) => splitter.semigroup.plus(old, stats)
+          case None => stats
         }
+        leafStats += feature -> combined
+        treeStats.update(leafIndex, leafStats)
+      }
+    }
```
striation: Since you are using this imperatively, I think it would be more efficient to do this imperatively (you wouldn't need to call `.toList` internally everywhere):

```scala
treeMap.foreach { case (treeIndex, tree) =>
  allStats.get(treeIndex).foreach { treeStats =>
    val times = sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex)
    (1 to times).foreach { i =>
      tree.leafFor(instance.features).foreach { case (leafIndex, target, annotation) =>
        if (stopper.shouldSplit(target) && treeStats.containsKey(leafIndex)) {
          features.foreach { case (feature, stats) =>
            if sampler.includeFeature(feature, treeIndex, leafIndex) {
              var leafStats = treeStats.get(leafIndex).getOrElse(Map[K,splitter.S]())
              val combined = leafStats.get(feature) match {
                case Some(old) => splitter.semigroup.plus(old, stats)
                case None => stats
              }
              leafStats += feature -> combined
              treeStats.update(leafIndex, leafStats)
            }
          }
        }
      }
    }
  }
}
```

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -6,81 +6,145 @@ import com.twitter.algebird._
...
+      for (
+        (treeIndex, tree) <- treeMap.toList;
+        i <- 1.to(sampler.timesInTrainingSet(instance.id, instance.timestamp, treeIndex)).toList;
+        leafIndex <- tree.leafIndexFor(instance.features).toList
+      ) {
+        val treeTargets = targets(treeIndex)
+        val old = treeTargets.getOrElse(leafIndex, Monoid.zero[T])
+        val combined = Monoid.plus(instance.target, old)
+        targets += treeIndex -> (treeTargets + (leafIndex -> combined))
+      }
```
striation: Same comment as above -- I think converting this to `.foreach` would be more efficient.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -6,81 +6,145 @@ import com.twitter.algebird._
...
-      val useTrees = trees.zipWithIndex.filter {
-        case (tree, i) =>
-          sampler.includeInValidationSet(instance.id, instance.timestamp, i)
-      }.map { _._1 }
-      if(useTrees.isEmpty)
-        None
-      else {
-        val prediction = voter.predict(useTrees, instance.features)
-        Some(error.create(instance.target, prediction))
+    var output: Option[E] = None
```
striation: Do we expect that we won't have a monoid for error? It seems like using `Monoid[E]` here would be preferable unless that was a non-starter.

brushfire-core/src/main/scala/com/stripe/brushfire/local/Example.scala
```diff
@@ -9,31 +9,25 @@ import AnnotatedTree.{AnnotatedTreeTraversal, fullBinaryTreeOpsForAnnotatedTree}
 object Example extends Defaults {
 
   def main(args: Array[String]) {
-    val cols = args.toList
-  
-    val trainingData = io.Source.stdin.getLines.map { line =>
+    val path = args.head
+    val cols = args.tail.toList
```
striation: Really small thing here but you could say:

```scala
val path :: cols = args.toList
```

avibryant:
Per IRL conversation, I'm going to merge this and @striation might do a follow-up PR with his mutable optimizations.

PR #91, opened 2016-04-16
Factor out trees to support evaluation without training (issue #51)

johnynek:
This factors out a `brushfire-trees` module. It had to include `Error` unlike #51 wanted, unless we want to refactor some, but it only depends on algebird and bonsai.

Also it breaks the style of having core types in `Brushfire.scala` and the a Foos.scala file. Instead, I renamed those to just Foo.scala and moved the type to the top.


johnynek:
?r @avibryant

johnynek:
PS: can you give me commit rights on the repo? Had to pull from my account.

brushfire-tree/src/main/scala/com/stripe/brushfire/Split.scala
```diff
@@ -0,0 +1,18 @@
+package com.stripe.brushfire
+
+/** Candidate split for a tree node */
+case class Split[V, T](predicate: Predicate[V], leftDistribution: T, rightDistribution: T) {
```
avibryant: Split is only used in training and should go there.

brushfire-tree/src/main/scala/com/stripe/brushfire/Dispatched.scala
```diff
@@ -0,0 +1,34 @@
...
+      case _ => sys.error("Values cannot be compared: " + (left, right))
+    }
+  }
+
+  def ordinal[A](a: A) = Ordinal(a)
+  def nominal[B](b: B) = Nominal(b)
+  def continuous[C](c: C) = Continuous(c)
+  def sparse[D](d: D) = Sparse(d)
+
+  def wrapSplits[X, T, A: Ordering, B, C: Ordering, D](splits: Iterable[Split[X, T]])(fn: X => Dispatched[A, B, C, D]) = splits.map(_.map(fn))
```
avibryant: you'll have to move this into the training package

brushfire-tree/src/main/scala/com/stripe/brushfire/Error.scala
```diff
@@ -1,7 +1,20 @@
 package com.stripe.brushfire
 
-import com.twitter.algebird._
-
+import com.twitter.algebird.{AveragedValue, Group, Monoid, Semigroup}
+
+/** Computes some kind of error by comparing the trees' predictions to the validation set */
+trait Error[T, P, E] {
```
avibryant: I think you probably have this in here because pruning requires it, but pruning is really a training thing, and we should move this to training and then extract pruning from tree. (Or, IMO it even would be fine to temporarily lose the ability to prune and file an issue to bring it back).

avibryant:
* You should have bits now.
* See comments inline.
* Feels a bit strange to call the training package `brushfire-core`, since if anything `brushfire-tree` is more core in the dependency graph. Can we rename it to `brushfire-training`?

brushfire-tree/src/main/scala/com/stripe/brushfire/Error.scala
avibryant: Though TBH I can see an argument for including pruning and error here, also - basically, even though they are part of the training phase, they are likely to be things we want, both in interface and implementation, for *any* training approach, vs all of the splitter/evaluator/etc machinery which are very specific to the particular approach we're taking. So, fine, leave it :)

johnynek:
Thanks for the review Avi. One question that came up while doing this was why (other than see the code a bit)?

Operationally, splitting packages brings a (small) cost to consuming, but the benefit is if some consumers can have thinner dependencies by avoiding the transitives of some parts. Here, I think that is really only the serialization stuff (which could be in a package like `brushfire-serialization` that sits between `tree` and `training`.

But note, we can't trim any deps by moving `Error` or `Split` so, I'm not 100% sure it is worth it to refactor. What do you think?

I can definitely do the refactor you suggest, but so far we are source and binary compatible, so it won't break. The refactor will break both, are we sure we want that? Don't want to do more work and then we get cold feet and don't merge.

What do you think of `brushfire-serialization` in between `tree` and `training`?

johnynek:
I'll split serialization out in a follow up. This is big enough

johnynek:
actually went ahead and split out serialization since it has a ton of dependencies and that gets you a nice win if you care about that.

Actually, `brushfire-scalding` only has an example dependency on `brushfire-serialization` so we could even break that by making the example something we put somewhere else.

avibryant:
LGTM.

PR #93, opened 2016-06-24
Add a CrossEntropyError class.

erik-stripe:
This PR gives us another way to evaluate how well predictions do
against the actual known distribution. The iris example has been
ported to demonstrated this method in practice.

There is also a small refactoring of the local trainer's validate
method, and some small refactors of other error classes.

### DIFF
--- a/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
+++ b/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
@@ -45,7 +45,7 @@ case class Trainer[K: Ordering, V: Ordering, T: Monoid](
     copy(samplerExecution = newExecution)
   }
 
-  def tee(fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[_]): Trainer[K, V, T] = {
+  def tee[A](fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[A]): Trainer[K, V, T] = {
     val newExecution = treeExecution
       .flatMap { trees =>
         Execution.zip(trainingDataExecution, samplerExecution, trees.toIterableExecution)
--- a/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
+++ b/brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
@@ -45,7 +45,7 @@ case class Trainer[K: Ordering, V: Ordering, T: Monoid](
     copy(samplerExecution = newExecution)
   }
 
-  def tee[A](fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[A]): Trainer[K, V, T] = {
+  def tee(fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[_]): Trainer[K, V, T] = {
     val newExecution = treeExecution
       .flatMap { trees =>
         Execution.zip(trainingDataExecution, samplerExecution, trees.toIterableExecution)
--- a/brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
+++ b/brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
@@ -27,12 +27,14 @@ object CrossEntropyExample extends Defaults {
     val totalNormalized: Map[String, Double] = CrossEntropyError.normalize(totalData)
     val totalEntropy: Double = CrossEntropyError.entropy(totalNormalized)
 
-    def relativeEntropy(xn: (Double, Long)): Double = {
-      val (x, n) = xn
+    def relativeEntropy(xn: AveragedValue): Double = {
+      val x = xn.value
+      val n = xn.count
       (totalEntropy - (x / n)) / totalEntropy
     }
 
-    val t0 = Trainer(trainingData, KFoldSampler(4)).updateTargets
+    //val t0 = Trainer(trainingData, KFoldSampler(4)).updateTargets
+    val t0 = Trainer(trainingData, RFSampler(10, 0.3)).updateTargets
 
     (0 until 10).foldLeft(t0) { (trainer, i) =>
       println(i)
--- a/brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
+++ b/brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
@@ -1,0 +1,44 @@
+package com.stripe.brushfire.local
+
+import com.stripe.bonsai._
+import com.stripe.brushfire._
+import com.twitter.algebird._
+
+import AnnotatedTree.{AnnotatedTreeTraversal, fullBinaryTreeOpsForAnnotatedTree}
+
+object CrossEntropyExample extends Defaults {
+
+  def main(args: Array[String]) {
+    val path = args.head
+    val cols = args.tail.toList
+
+    // T = Map[String, Long]
+    // P = String
+    // E = Double
+
+    val trainingData = Lines(path).map { line =>
+      val parts = line.split(",").reverse.toList
+      val label = parts.head
+      val values = parts.tail.map { s => s.toDouble }
+      Instance(line, 0L, Map(cols.zip(values): _*), Map(label -> 1L))
+    }.toIterable
+
+    val totalData: Map[String, Long] = Monoid.sum(trainingData.iterator.map(_.target))
+    val totalNormalized: Map[String, Double] = CrossEntropyError.normalize(totalData)
+    val totalEntropy: Double = CrossEntropyError.entropy(totalNormalized)
+
+    def relativeEntropy(xn: (Double, Long)): Double = {
+      val (x, n) = xn
+      (totalEntropy - (x / n)) / totalEntropy
+    }
+
+    val t0 = Trainer(trainingData, KFoldSampler(4)).updateTargets
+
+    (0 until 10).foldLeft(t0) { (trainer, i) =>
+      println(i)
+      val error = trainer.validate(CrossEntropyError())
+      println(error.map(relativeEntropy))
+      trainer.expand(1)
+    }
+  }
+}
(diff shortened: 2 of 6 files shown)


striation:
Review by @tixxit, @avibryant, and/or @johnynek.

striation:
There are some problems here -- please wait to merge until I fix them (Travis should notice them too).

brushfire-scalding/src/main/scala/com/stripe/brushfire/scalding/Trainer.scala
```diff
@@ -45,7 +45,7 @@ case class Trainer[K: Ordering, V: Ordering, T: Monoid](
     copy(samplerExecution = newExecution)
   }
 
-  def tee[A](fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[A]): Trainer[K, V, T] = {
+  def tee(fn: ((TypedPipe[Instance[K, V, T]], Sampler[K], Iterable[(Int, Tree[K, V, T])])) => Execution[_]): Trainer[K, V, T] = {
```
avibryant: Just in the interests of increasing my Scala knowledge: what difference does this make?

brushfire-training/src/main/scala/com/stripe/brushfire/local/Trainer.scala
```diff
@@ -126,26 +126,21 @@ case class Trainer[K: Ordering, V: Ordering, T: Monoid](
...
+      val features = instance.features
+
+      val predictions = treeMap.iterator
+        .filter { case (treeIndex, _) =>
+          sampler.includeInValidationSet(id, timestamp, treeIndex)
+        }.flatMap { case (_, tree) =>
+          tree.targetFor(features)
+        }.toVector
+
+      error.create(instance.target, voter.combine(predictions))
```
avibryant: I think this is going to give you the wrong answer when `predictions` is empty (eg because this instance is not in the validation set). Who knows what `voter.combine(Vector.empty)` is going to produce, but it's totally possible (even, likely) that it will produce a non-zero error, which means we're accumulating all kinds of extra error for stuff that should have been filtered out.

brushfire-tree/src/main/scala/com/stripe/brushfire/Error.scala
```diff
@@ -27,13 +28,10 @@ trait FrequencyError[L, M, E] extends Error[Map[L, M], Map[L, Double], E] {
...
 
-  def create(actual: Map[L, M], predicted: Map[L, Double]) = {
-    if (predicted.isEmpty)
-      monoid.zero
-    else {
-      monoid.sum(actual.map { case (label, count) => error(label, count, predicted) })
-    }
-  }
+  def create(actual: Map[L, M], predicted: Map[L, Double]): E =
+    monoid.sum(actual.iterator.map { case (label, count) =>
```
avibryant: you're removing the protection of the `predicted.isEmpty` case, which means the `maxBy` in `error` is going to throw if that's true, right?

brushfire-training/src/main/scala/com/stripe/brushfire/local/Trainer.scala
striation: I think we had convinced ourselves that there wouldn't be error in this case. But it's easy to add that special-case back in if need be.

avibryant: I don't see how you can know that? There's no obvious law that Voter and Error need to conform to which would lead to

`error.create(t, voter.combine(Vector.empty[T]]) == error.monoid.zero`

brushfire-tree/src/main/scala/com/stripe/brushfire/Error.scala
johnynek: no. `monoid.zero` is already the return for `monoid.sum` when you are empty, right?

johnynek: I misunderstood... Sorry.

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
```diff
@@ -0,0 +1,44 @@
...
+      val label = parts.head
+      val values = parts.tail.map { s => s.toDouble }
+      Instance(line, 0L, Map(cols.zip(values): _*), Map(label -> 1L))
+    }.toIterable
+
+    val totalData: Map[String, Long] = Monoid.sum(trainingData.iterator.map(_.target))
+    val totalNormalized: Map[String, Double] = CrossEntropyError.normalize(totalData)
+    val totalEntropy: Double = CrossEntropyError.entropy(totalNormalized)
+
+    def relativeEntropy(xn: (Double, Long)): Double = {
```
johnynek: this should really be called `normalizedInformation` or something (how much of the total mutual information have we learned).

After 10 runs we got to like 60% or something as I recall.

brushfire-tree/src/main/scala/com/stripe/brushfire/CrossEntropyError.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def create(actual: Map[L, R], predicted: Map[L, Double]): (Double, Long) = {
+    val p = actual.iterator.map { case (k, r) => (k, num.toDouble(r)) }.toMap
+    CrossEntropyError.crossEntropyError(p, predicted)
+  }
+}
+
+object CrossEntropyError {
+
+  def normalize[L, R](p: Map[L, R])(implicit num: Numeric[R]): Map[L, Double] = {
```
johnynek: might call this `l1Normalize` maybe?

brushfire-tree/src/main/scala/com/stripe/brushfire/CrossEntropyError.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def entropy[L](p: Map[L, Double]): Double =
+    -p.iterator.map { case (_, pi) => plogp(pi) }.sum
+
+  // we assumed the distribution was q, but it actually follows distribution p.
+  def crossEntropyError[L](p0: Map[L, Double], q0: Map[L, Double]): (Double, Long) = {
+    val ks = p0.keySet | q0.keySet
+    val p = normalize(p0).withDefault(_ => 0.0)
+    val q = clamp(ks, q0)
+    (-ks.iterator.map { k => p(k) * log2(q(k)) }.sum, 1L)
```
avibryant: I don't understand why we're normalizing `p` and then always producing a weight of `1L` here. (I assume the `Long` element in the result tuple is a weight/count). I would have expected that to be `p0.values.sum`.

Or, better (I think) would be to use `FrequencyError` and have this method follow the signature of that `error` method, which just computes the error for a single label in the distribution (like, a single (L,M) label/count pair).

avibryant:
Having other people extend Brushfire really makes me feel the lack of having written tests. This makes me feel bad but is good for improving the code.

Here's a law that I think we want to apply for all `Error` instances that I think this currently fails:

```scala
error.semigroup.plus(error.create(a1, p), error.create(a2, p)) == error.create(Semigroup.plus(a1,a2),p)
```

Please note, this does *not* hold for predictions, that is:

```scala
error.semigroup.plus(error.create(a, p1), error.create(a, p2)) != error.create(a, Semigroup.plus(p1,p2))
```

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
```diff
@@ -0,0 +1,44 @@
...
+    // E = Double
+
+    val trainingData = Lines(path).map { line =>
+      val parts = line.split(",").reverse.toList
+      val label = parts.head
+      val values = parts.tail.map { s => s.toDouble }
+      Instance(line, 0L, Map(cols.zip(values): _*), Map(label -> 1L))
+    }.toIterable
+
+    val totalData: Map[String, Long] = Monoid.sum(trainingData.iterator.map(_.target))
```
avibryant: It seems wrong to me that we are using the full training set, in any way, in an error computation; I feel like we should only be basing the error computation on the validation set.

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
avibryant: If relativeEntropy is the metric we really care about, can we build that into the `CrossEntropyError`? It seems like the only thing we need to keep track of to compute this is the total distribution of actuals, which would be easy to include in the error monoid.

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
```diff
@@ -0,0 +1,44 @@
...
+    val totalNormalized: Map[String, Double] = CrossEntropyError.normalize(totalData)
+    val totalEntropy: Double = CrossEntropyError.entropy(totalNormalized)
+
+    def relativeEntropy(xn: (Double, Long)): Double = {
+      val (x, n) = xn
+      (totalEntropy - (x / n)) / totalEntropy
+    }
+
+    val t0 = Trainer(trainingData, KFoldSampler(4)).updateTargets
```
avibryant: We should try `RFSampler(10)` here and see how much better the results are.

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
johnynek: yeah, that's fine. We just need an estimate of the "true" entropy. We should use the exact same set that we are measuring the error on below.

I guess we could fit it into the semigroup approach by aggregating the total label distribution as we go. (in that way, by the way), it seems like `Error` should really be a semigroup and a function (maybe a full aggregator).

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
```diff
@@ -0,0 +1,44 @@
...
+
+    (0 until 10).foldLeft(t0) { (trainer, i) =>
+      println(i)
```
tixxit: --

brushfire-training/src/main/scala/com/stripe/brushfire/local/CrossEntropyExample.scala
avibryant: Agreed that we should do that (that's what I meant in https://github.com/stripe/brushfire/pull/93/files/6aa9b7fe7adaaf9dbbce7f79abd767e56c4f2ecd#r68490840). I take your point about `Error` needing a `present` function. `Error[T,P,E]` could equally be `Aggregator[(T,P),E,E2]` where `E2` is the thing you actually care about. Or you could decide you just cared about having an `Ordering[E]` (which comes up a lot in practice) and which would do the transformation to `Double` or whatever internally.

PR #95, opened 2016-07-21
Quick Start example in README. 

eightysteele:
This PR adds some love to the Quick Start example in README. :)
- Fixed an issue where the SNAPSHOT version in `example/iris` didn't match what was in `version.sbt`. 
- Fixed an issue where `org.apache.hadoop` was excluded from runtime by sbt-assembly, causing a `ClassNotFoundException` to be thrown by `example/iris`. Basically the `hadoopClient` in `Deps.scala` was "provided" so it wasn't getting included in `target/scala-2.11/brushfire-scalding-0.7.5-SNAPSHOT-jar-with-dependencies.jar` during builds.


avibryant:
This is great, thanks! I think the reason that `hadoopClient` was listed as `provided` is that when you are submitting the assembly jar to a hadoop cluster, the hadoop jars are indeed provided in that execution environment, and it can be problematic to duplicate them. But obviously that's not the case when running locally. I'm not sure what the best way to resolve this is.

eightysteele:
Good catch! 

The sbt-assembly [docs](https://github.com/sbt/sbt-assembly#excluding-jars-and-files) offers a clue on how to resolve this.

If we add this to `brushfire-scalding/build.sbt`:

```scala
run in Compile <<= Defaults.runTask(fullClasspath in Compile, mainClass in (Compile, run), runner in (Compile, run))

runMain in Compile <<= Defaults.runMainTask(fullClasspath in Compile, runner in(Compile, run))
```

Then we can run it locally (with all the dependencies) like this:

```bash
$ sbt "brushfireScalding/runMain com.twitter.scalding.Tool com.stripe.brushfire.scalding.IrisJob --local --input example/iris.data --output example/iris.output"
```

Boom! 

I wrapped the above command in a new script called `quick-start`, moved `hadoopClient` back to `provided`, and updated the README with examples of running locally and on the cluster.

There might be an even BETTER way to resolve this. Happy to pivot. Tell me what you think. :)

PR #98, opened 2016-11-11
Experimental iterator for depth-first tree traversals

jhoon-stripe:
I don't intend to merge this pull request. I'm putting this up here to demonstrate a method for obtaining a functional iterator for trees. This was a way for me to get a bit more familiar with dependent types (since we need to work with the type member of `FullBinaryTreeOps`) and generally play with Scala data structures.

At its cores, this PR introduces a wrapper around a stack of tree nodes, which allows us to iterate through the tree's nodes depth-first.

cc @rob-stripe  @sritchie-stripe @thomas-stripe @erik-stripe @oscar-stripe @kelley-stripe @avi-stripe


erik-stripe:
Nice! Glad you got the path-dependent types working.

As an additional challenge, you could try implementing `streamNode` directly with `foldNode`.

